PR #76, opened 2019-01-08
Write bayesian LS log

eivantsov:
It will be nice to have Bayesian Language Server log file to debug problems


dmytro-ndp:
We have dedicated location for workspace logs: */workspace_logs*.
What about putting Bayesian LS log file into the directory */workspace_logs/ls-bayesian* directory?

PR #134, opened 2019-04-09
wrap call to...

nickboldt:
wrap call to get-sources-rhpkg-container-build in a try block; set buildstatus UNSTABLE if cannot trigger

Change-Id: I2686e9988d4d84a0fb1087b192818fe50ae1f0c8


nickboldt:
masterPR

PR #153, opened 2019-05-10
Make sure wsagent runs on openjdk11

eivantsov:
Adding a custom setenv.sh and bind+activation dependencies.


assembly/assembly-wsagent-server/src/assembly/setenv.sh
```diff
@@ -0,0 +1,44 @@
...
+                                             -Dcom.sun.management.jmxremote.authenticate=false \
+                                             -Dche.local.conf.dir=${CHE_LOCAL_CONF_DIR}"
+
+#Class path
+[ -z "${CLASSPATH}" ]  && CLASSPATH="${CATALINA_HOME}/conf/:${JAVA_HOME}/lib/tools.jar"
+
+
+# On java9 runtime, enable activation and JAXB API
+#JAVA_VERSION=$(java -version 2>&1 | head -n 1 | cut -d'"' -f2 | cut -d'.' -f1 | sed 's/[^0-9]*//g')
+#[ "$JAVA_VERSION" -ge 9 ] && JAVA_OPTS="$JAVA_OPTS --add-modules java.activation --add-modules java.xml.bind"
```
nickboldt: we could pull this from upstream and just delete lines 37-39 (or comment out line 39) with an ant regex replacement.

assembly/assembly-wsagent-war/pom.xml
```diff
@@ -24,6 +24,26 @@
     <packaging>war</packaging>
     <name>CodeReady Workspaces :: Workspace Agent</name>
     <dependencies>
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.2.11</version>
```
nickboldt: If this is inheriting from <dependencyManagement> we don't need versions here.

eivantsov: Good catch. Fixed.

nickboldt:
merged

PR #157, opened 2019-05-23
CRW-256 fixed in upstream Che...

nickboldt:
CRW-256 fixed in upstream Che 6.19.5

Change-Id: I33641951d26523c697122e5d976f4937a4c69512


nickboldt:
Forgot to merge for 1.2, will pick this up for 1.2.1

PR #176, opened 2019-10-08
 web-nodejs-simple devfile

JPinkney:
This PR brings in the web-nodejs-simple devfile.

This devfile differs from the devfile in the che-devfile-registry in 2 ways:
- Uses registry.redhat.io/codeready-workspaces/stacks-node-rhel8 which includes node 10
- nodemon isn't available in the current version of the image so it has replaced all references of nodemon with node (though I believe @nickboldt is working on this so that this image has nodemon)

Signed-off-by: Josh Pinkney <joshpinkney@gmail.com>

### DIFF
--- a/dependencies/che-devfile-registry/devfiles/web-nodejs-simple/devfile.yaml
+++ b/dependencies/che-devfile-registry/devfiles/web-nodejs-simple/devfile.yaml
@@ -1,0 +1,66 @@
+---
+apiVersion: 1.0.0
+metadata:
+  generateName: nodejs-
+projects:
+  - name: nodejs-web-app
+    source:
+      type: git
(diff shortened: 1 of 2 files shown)


nickboldt:
New container build is done. Try quay.io/crw/stacks-node-rhel8:2.0-3 or newer and let me know if nodemon works for you in there.

JPinkney:
@nickboldt Everything works as expected with that image :) :+1: Do you want me to push the changes using the quay image and then we change it to the red hat registry image when CRW2 is released?

nickboldt:
Yes, please update PR to use nodemon and latest quay image. We can flip these over to use reg.rh.io later when we're closer to GA.

JPinkney:
@nickboldt Done!

PR #178, opened 2019-10-09
Add dotnet, java maven and jboss devfiles

svor:

Adds several devfiles:
-  dotnet with [dotnet-web-simple](https://github.com/che-samples/dotnet-web-simple) application
-  java-maven with [console-java-simple](https://github.com/che-samples/console-java-simple.git) application
-  java-eap-maven with [kitchensink-example](https://github.com/che-samples/kitchensink-example.git)
- java-eap-thorntail with [rest-http-redhat](https://github.com/thorntail-examples/rest-http-redhat) application

### DIFF
--- a/dependencies/che-devfile-registry/devfiles/dotnet/devfile.yaml
+++ b/dependencies/che-devfile-registry/devfiles/dotnet/devfile.yaml
@@ -1,0 +1,43 @@
+---
+apiVersion: 1.0.0
+metadata:
+  generateName: dotnet-
+projects:
+  -
+    name: dotnet-web-simple
+    source:
+      type: git
+      location: "https://github.com/che-samples/dotnet-web-simple"
+components:
+  -
+    type: chePlugin
+    alias: omnisharp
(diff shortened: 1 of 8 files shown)


svor:
@nickboldt it is ready to review

svor:
@nickboldt it is ready to review

nickboldt:
If you update your PR to the latest in master you can then run a build:

`docker build --no-cache -t quay.io/eclipse/che-devfile-registry:nightly -f build/dockerfiles/rhel.Dockerfile .`

And then use that devfile registry in your local Che to see how it behaves. 

Meanwhile I notice that the java-maven config is different from the upstream one:

image

Do we need the same JAVA_OPTS and JAVA_TOOL_OPTIONS? or are those only for JDK 11? 

Also are we sure redhat/java/latest will always be JDK 8? maybe it's better to use `redhat/java8/latest` plugin instead?

svor:
Updated

PR #188, opened 2019-10-24
Add support for caching prebuilt vsix extensions

mkuznyetsov:
This PR adds the ability to provide during the build of plugin registry an archive with .vsix extensions, so that during the caching artifacts stage it would reuse them, instead of downloading by extension URL. For this to work, an archive `vsix.tar.gz` must be placed in a `/v3` directory before the build of plugin registry, and it must contain a plugin in a folder structure, following the extension URL path (including the `vsix` root directory of tarball)

For example, a `vsix.tar.gz` with a following extension in it:
```
vsix/marketplace.visualstudio.com/_apis/public/gallery/publishers/ms-vscode/vsextensions/node-debug2/1.33.0/vspackage/node-debug2-1.33.0.vsix
```

Would have this extension reused for plugins with following extension URL
```
meta.yaml:
...
  extensions:
  - https://marketplace.visualstudio.com/_apis/public/gallery/publishers/ms-vscode/vsextensions/node-debug2/1.33.0/vspackage
```

So this feature would allow avoiding attempting to download of any plugins, that would rather need to be rebuilt and provided by us.

PR comes with new Jenkins Jobs
- Jenkinsfile-all - main job, that fetches plugin registry, looks up the vsix. extensions URL (currently only those that are hosted on "marketplace.visualstudio.com", builds them and collects in a tarball vsix.tar.gz
- Jenkinsfile-single - sub-job, that packages a vsix extensions from given repository

### DIFF
--- a/dependencies/che-plugin-registry/Jenkinsfile-all
+++ b/dependencies/che-plugin-registry/Jenkinsfile-all
@@ -1,0 +1,71 @@
+#!/usr/bin/env groovy
+
+// build params
+//BranchToBuildCRW - branch for crw plugin registry
+
+timeout(120) {
+    node {
+        cleanWs()
+        stage 'build extensions'
+
+        // key value map for build jobs, where key is extension location in the registry
+        def jobs = [:]
+
+        checkout([$class: 'GitSCM',
+            branches: [[name: "${branchToBuildCRW}"]],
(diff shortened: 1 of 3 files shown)


nickboldt:
PR was in draft mode, so couldn't merge. Instead I cherrypicked your commit, so we can close this. Thanks!!

PR #200, opened 2019-11-20
Fix launching remote plugins according to use latest che-plugin-broker.

AndrienkoAleksandr:
### Referenced issue:
https://issues.jboss.org/browse/CRW-493

### What does this pull request do
Change plugin meta.yaml#spec#containers#container#args(it is Dockerfile#CMD) to start remote plugin endpoint binary.  

Notice:
We don't touch meta.yaml#spec#containers#container#command (it is Dockerfile#ENTRYPOINT), so entrypoint.sh scripts defined in the Dockerfile#ENTRYPOINT should work fine.

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>

### DIFF
--- a/dependencies/che-plugin-registry/v3/plugins/che-incubator/cpptools/0.1/meta.yaml
+++ b/dependencies/che-plugin-registry/v3/plugins/che-incubator/cpptools/0.1/meta.yaml
@@ -15,6 +15,10 @@ spec:
     - image: "quay.io/crw/stacks-cpp-rhel8:2.0"
       name: cpp-plugins
       memoryLimit: '512Mi'
+      args:
+        - sh
+        - -c
+        - ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}
   extensions:
     - https://github.com/che-incubator/che-cpptools/releases/download/v1.0/cdt-gdb-vscode-0.0.90.vsix
     - https://github.com/che-incubator/che-cpptools/releases/download/v1.0/cdt-vscode-0.0.7.vsix
--- a/dependencies/che-plugin-registry/v3/plugins/che-incubator/typescript/1.35.1/meta.yaml
+++ b/dependencies/che-plugin-registry/v3/plugins/che-incubator/typescript/1.35.1/meta.yaml
@@ -15,5 +15,9 @@ spec:
   - image: "quay.io/crw/stacks-node-rhel8:2.0"
     name: vscode-typescript
     memoryLimit: '512Mi'
+    args:
+      - sh
+      - -c
+      - ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}
   extensions:
   - https://github.com/che-incubator/ms-code.typescript/releases/download/v1.35.1/che-typescript-language-1.35.1.vsix
--- a/dependencies/che-plugin-registry/v3/plugins/ms-python/python/2019.5.18875/meta.yaml
+++ b/dependencies/che-plugin-registry/v3/plugins/ms-python/python/2019.5.18875/meta.yaml
@@ -16,5 +16,9 @@ spec:
   - image: "quay.io/crw/stacks-python-rhel8:2.0"
     name: vscode-python
     memoryLimit: '512Mi'
+    args:
+      - sh
+      - -c
+      - ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}
   extensions:
   - https://github.com/Microsoft/vscode-python/releases/download/2019.5.18875/ms-python-release.vsix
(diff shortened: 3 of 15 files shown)


AndrienkoAleksandr:
@nickboldt could you review please?

AndrienkoAleksandr:
Also I have one concern about kubernetes image https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-plugin-registry/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.0.4/meta.yaml#L15. In this image we have entrypoint.sh script which should apply permission fix using template files from /home/theia folder. But this folder is empty and that's why permission fix fail. @nickboldt Could you take a look please?

nickboldt:
Merged. @AndrienkoAleksandr can you throw me a PR for fixing http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-plugin-kubernetes/tree/?h=crw-2.0-rhel-8 and http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-plugin-openshift/tree/?h=crw-2.0-rhel-8 ?

nickboldt:
> In this image we have entrypoint.sh script which should apply permission fix using template files from /home/theia folder. But this folder is empty and that's why permission fix fail.

I've fixed the Dockerfile so that the permissions are set up correctly from the template files.

Kube plugin build is now green, including CVP tests.

However I can't get the odo executable to run inside an OCP node.

I've updated our build (see links to Dockerfile below) to use odo 1.0.1, and it just passed in Brew. But the CVP tests still fail with this error:

> ?  user: unknown userid 1001850000\ncommand terminated with exit code 1

The test is simply trying to do this:

$ oc exec kube-cmd-test-1-skwzs -- odo version

Full log:

https://jenkins-cvp-5bace840bed8bd6ee81a91c4.cloud.paas.psi.redhat.com/blue/rest/organizations/jenkins/pipelines/cvp-product-test/runs/9/log/?start=0

Any advice for how to make odo run inside an openshift pod? it runs during the brew build itself (I can do `odo version`) but ... while I can run `oc`, `helm` or `kubectl` as a CVP test, `odo` does NOT run due to the 'unknown userid` thing above.

Is this an issue with odo, or an issue with the way the container is built?

Here's the CVP test source:

http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-plugin-openshift/tree/tests/basic-test.yaml?h=crw-2.0-rhel-8#n45

And here's the Dockerfile:

http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-plugin-openshift/tree/Dockerfile?h=crw-2.0-rhel-8

Anything you can suggest would be great.

AndrienkoAleksandr:
Hello. I saw your Docker file change and they looks good. But yes issue with odo looks strange. I will try to understand what's wrong.

rhopp:
@AndrienkoAleksandr I think the entrypoint is wrong.
```
if ! grep -Fq "${USER_ID}" /etc/passwd; then
    # current user is an arbitrary
    # user (its uid is not in the
    # container /etc/passwd). Let's fix that
    cat ${HOME}/passwd.template | \
    sed "s/\${USER_ID}/${USER_ID}/g" | \
    sed "s/\${GROUP_ID}/${GROUP_ID}/g" | \
    sed "s/\${HOME}/\/home\/theia/g" > /etc/passwd

    cat ${HOME}/group.template | \
    sed "s/\${USER_ID}/${USER_ID}/g" | \
    sed "s/\${GROUP_ID}/${GROUP_ID}/g" | \
    sed "s/\${HOME}/\/home\/theia/g" > /etc/group
fi
```
Seems to me, that the sed is trying to replace placeholders in $HOME/passwd.template, but there are no such placeholders:
```
$ cat $HOME/passwd.template 
root:x:0:0:root:/root:/bin/bash
bin:x:1:1:bin:/bin:/sbin/nologin
daemon:x:2:2:daemon:/sbin:/sbin/nologin
adm:x:3:4:adm:/var/adm:/sbin/nologin
lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
sync:x:5:0:sync:/sbin:/bin/sync
shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
halt:x:7:0:halt:/sbin:/sbin/halt
mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
operator:x:11:0:operator:/root:/sbin/nologin
games:x:12:100:games:/usr/games:/sbin/nologin
ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
nobody:x:65534:65534:Kernel Overflow User:/:/sbin/nologin
```
resulting in unknown user:
```
$ whoami
whoami: cannot find name for user ID 1000620000
```
And according to https://github.com/openshift/odo/issues/2380, odo cannot deal with that.

PR #201, opened 2019-11-20
Workaround for CRW-494 - disable pushing logs to server in init broker

amisevsk:
This PR is a workaround for https://issues.jboss.org/browse/CRW-494 pending merge of PRs https://github.com/eclipse/che/pull/15263, https://github.com/eclipse/che/pull/15264. 

During the docker build of the init container, we comment out the lines that attempt to connect to the server via websocket. Since upstream Che 7.3.2 does not provision proxy settings for the init container, connecting fails and this blocks workspace start. The only impact of removing these lines is that workspace start does not see the log message
```
Cleaning /plugins dir
```

### DIFF
--- a/dependencies/che-pluginbroker/build/init/rhel.Dockerfile
+++ b/dependencies/che-pluginbroker/build/init/rhel.Dockerfile
@@ -16,6 +16,13 @@ ENV PATH=/opt/rh/go-toolset-1.11/root/usr/bin:$PATH \
 USER root
 WORKDIR /go/src/github.com/eclipse/che-plugin-broker/brokers/init/cmd/
 COPY . /go/src/github.com/eclipse/che-plugin-broker/
+# Workaround for CRW-494 - disable websocket communication in init broker
+RUN LN_START="$(grep -n 'cfg.DisablePushingToEndpoint' main.go | cut -f 1 -d ':')"; \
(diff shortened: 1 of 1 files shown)


amisevsk:
Just realized that this won't work if workspaces are ephemeral, since the regular broker is also provisioned as an init container.

dmytro-ndp:
There are workspaces which ephemeral by default 
- Spring Boot  https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-devfile-registry/devfiles/03_spring-boot-http-booster/devfile.yaml#L11
- Vert.x https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-devfile-registry/devfiles/03_vertx-http-booster/devfile.yaml#L11

ibuziuk:
@dmytro-ndp any reason to make them ephemeral by default on CRW ?

dmytro-ndp:
I am not aware about the reasons.

dmytro-ndp:
They are not ephemeral in upstream:
- https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/java-web-vertx/devfile.yaml
- https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/java-web-spring/devfile.yaml

nickboldt:
Rejected in favour of #202

PR #203, opened 2019-11-21
Remove ephemeral mode from spring boot and vert.x devfiles

amisevsk:


nickboldt:
+1 followup to #201 discussion.

PR #215, opened 2020-01-13
add java plugin 0.55.1; do NOT mark 0.50.0 deprecated

nickboldt:
add java plugin 0.55.1; mark 0.50.0 deprecated

Change-Id: Id1e54d478e8550704d8c5b85705b88b9d6d76759


nickboldt:
@fbricon WDYT? Also should we do the same in Che 7.8 upstream?

dependencies/che-plugin-registry/v3/plugins/redhat/java/0.55.1/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+firstPublicationDate: "2020-01-13"
+spec:
+  containers:
+  - image: "registry.redhat.io/codeready-workspaces/stacks-java-rhel8:2.0"
+    name: vscode-java
+    memoryLimit: "1500Mi"
+    args:
+      - sh
+      - -c
+      - ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}
```
amisevsk: More a question: is this required specifically because of the image being used? This format differs significantly from what's done [upstream](https://github.com/che-dockerfiles/che-sidecar-java/blob/f76ca45e727748660c94f651efeccdfda5ad7461/Dockerfile#L31-L32)

In general, passing `sh -c <something>` as args is strange, since the typical construct is
- `cmd`: some script, or similar, that ends in `exec $@`
- `args`: whatever needs to run

Does the plugin remote executable need to be run by `sh`?

nickboldt: That was added here https://github.com/redhat-developer/codeready-workspaces/commit/bbfd4014ac8d6de6b34137da81d3fff917ee6309  by @AndrienkoAleksandr  but not sure what it does / why it's needed.

AndrienkoAleksandr: It is REQUIRED if we want to use image without ANY MODIFICATION to start remote endpoint.  See more, pr docs https://github.com/eclipse/che-docs/pull/1022

fbricon:
All I can say is it's better to use the latest version of vscode-java. Moar perf, less bugs.

honest question: Why isn't there an image for Java 11 support?

dependencies/che-plugin-registry/v3/plugins/redhat/java/0.55.1/meta.yaml
amisevsk: My question is mostly about the `sh -c` part; if the entrypoint is a wrapper script, is `sh -c` necessary? We're basically calling
```
exec "sh -c ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}"
```
In the docs PR link, you mention that you can start the executable as a dockerfile's CMD -- usually in that case entrypoint is something like `exec $@` anyways, so it appears as though `sh -c` is not required (example 3 in the docs PR).

AndrienkoAleksandr: Yes, looks like we can skip it.

nickboldt: OK, removed them here across 17 files https://github.com/redhat-developer/codeready-workspaces/commit/fd5cc004b80c806aca5d313249eb4ce503db611e

nickboldt:
> honest question: Why isn't there an image for Java 11 support?

Because we ran out of time in CRW 2.0.

Here's the CRW 2.1 JDK 11 plugin:

https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-plugin-registry/v3/plugins/redhat/java11/0.50.0/meta.yaml

So, should I close this PR if we are not ready/willing to move to 0.55.1 version ?

dependencies/che-plugin-registry/v3/plugins/redhat/java/0.55.1/meta.yaml
amisevsk: Sorry I think I was unclear in my suggestion: I'm mostly talking of removing the `sh -c` part -- AFAIK the `${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}` is still needed.

PR #222, opened 2020-01-25
CRW-522 add Gradle devfile +...

nickboldt:
CRW-522 add Gradle devfile + icon

Change-Id: I9e531d21c046ab44180c7f4f57a4c408345b36ee


nickboldt:
For some reason I can't assign to or request review from @svor ... weird. Anyway here's a devfile based loosely on the CRW 7.7 one. LMK if it needs to be changed.

nickboldt:
Tried to import to CRW 2.1 deployed to OCP 4.1 and got:

```
Error: Failed to start the workspace: "Failure executing: POST at: https://172.30.0.1/apis/project.openshift.io/v1/projectrequests. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. projectrequests.project.openshift.io is forbidden: User "system:serviceaccount:nb-crw21:che" cannot create resource "projectrequests" in API group "project.openshift.io" at the cluster scope. The error may be caused by an expired token or changed password. Update Che server deployment with a new token or password."
```

But on che.openshift.io I could import https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/fbf0f74e642a634ae8d3d964144ad61860f9b258/dependencies/che-devfile-registry/devfiles/03_java11-gradle/devfile.yaml and start the workspace:

image

image

Strangely, neither the build or run tasks are present in the workspace on initial startup, but after running the build and the run, I now have:

image

So this devfile seems OK.

nickboldt:
> This warning could be fixed by changing testCompile with testImplementation in the sample in build.gradle file

See https://github.com/che-samples/console-java-simple/pull/16

PR #224, opened 2020-01-27
Fix Python plugin

mkuznyetsov:
Fix VS code Python plugin, by correctly building its extension

### DIFF
--- a/dependencies/che-plugin-registry/Jenkinsfile-single
+++ b/dependencies/che-plugin-registry/Jenkinsfile-single
@@ -73,6 +73,16 @@ def buildPhpDebug(branchToBuildPlugin, extensionFolder) {
     sh "cd ${extensionFolder} && sudo npm install -g vsce gulp && npm install && npm run build && vsce package"
 }
 
+def buildVscodePython(branchToBuildPlugin, extensionFolder) {
+    installNPM()
+    sh "sudo /usr/bin/python3 -m pip install --upgrade pip"
+    buildNumber = branchToBuildPlugin.substring(branchToBuildPlugin.lastIndexOf('.') + 1)
+    echo "${buildNumber}"
+    sh "cd ${extensionFolder} && sudo npm install -g vsce gulp && npm ci && gulp updateBuildNumber --buildNumber ${buildNumber} && rm -rf node_modules"
+    archiveSources(extensionFolder)
+    sh "cd ${extensionFolder} && npm ci && gulp installPythonLibs && gulp prePublishBundle && vsce package"
+}
+
 def buildDefault(extensionFolder) {
     archiveSources(extensionFolder)
     installNPM()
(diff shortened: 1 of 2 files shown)


dependencies/che-plugin-registry/Jenkinsfile-single
```diff
@@ -73,6 +73,16 @@ def buildPhpDebug(branchToBuildPlugin, extensionFolder) {
     sh "cd ${extensionFolder} && sudo npm install -g vsce gulp && npm install && npm run build && vsce package"
 }
 
+def buildVscodePython(branchToBuildPlugin, extensionFolder) {
+    installNPM()
+    sh "sudo /usr/bin/python3 -m pip install --upgrade pip"
```
nickboldt: probably don't need this any more as it's implemented in the config for the rhel7 nodes:

https://gitlab.cee.redhat.com/codeready-workspaces/cloud-slaves/blob/master/label/linux/pip.sh

But doing it again here won't hurt anything and makes it clear we need latest pip.

dependencies/che-plugin-registry/Jenkinsfile-single
```diff
@@ -73,6 +73,16 @@ def buildPhpDebug(branchToBuildPlugin, extensionFolder) {
...
+    buildNumber = branchToBuildPlugin.substring(branchToBuildPlugin.lastIndexOf('.') + 1)
+    echo "${buildNumber}"
+    sh "cd ${extensionFolder} && sudo npm install -g vsce gulp && npm ci && gulp updateBuildNumber --buildNumber ${buildNumber} && rm -rf node_modules"
+    archiveSources(extensionFolder)
```
nickboldt: you're archiving after building? 

for other extensions, you're archiving before the build. Why do it afterwards here?

dependencies/che-plugin-registry/Jenkinsfile-single
```diff
@@ -148,8 +158,10 @@ timeout(120) {
             buildEclipseCdt(extensionFolder)
         } else if (extensionPath.contains("https://github.com/llvm/llvm-project")) {
             buildClangVscode(extensionFolder)
+        } else if (extensionPath.contains("https://github.com/microsoft/vscode-python")) {
+            buildVscodePython(branchToBuildPlugin, extensionFolder)
         } else if (extensionPath.contains("https://github.com/microsoft/vscode-java-debug")) {
-            buildJavaExtension(extensionFolder, extensionFolder)
+            buildJavaExtension(branchToBuildPlugin, extensionFolder)
```
nickboldt: oops! good typo fix :)

dependencies/che-plugin-registry/Jenkinsfile-single
mkuznyetsov: that's not the build, but required things to update the build number (18875) into plugin's package.json (and resulting artifact name). Then in this command I would node_modules (to save some hundreds of MBs) and package my sources. Only after that I would "npm ci" again and perform the build.

PR #233, opened 2020-02-12
Add new Quarkus plugin version

mkuznyetsov:
Add new version of quarkus plugin, deprecate the older one


tsmaeder:
+1 on deleting stuff we never released. Also, upstream has latest versions of java and java-debug. +1 on taking those.

tsmaeder:
There is a 55.1 version if vscode-java, I guess the did a bugfix-version.

PR #235, opened 2020-02-17
Add netcoredbg plugin; Add debugger into .Net devfile

svor:
Add netcoredbg plugin; Add debugger into .Net devfile

cc @nickboldt


dependencies/che-devfile-registry/devfiles/05_dotnet/meta.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 displayName: ".NET"
-description: .NET 2.1 stack with .NET Core SDK and Runtime
+description: .NET 3.1 stack with .NET Core SDK and Runtime
 tags: [".NET", "C#"]
```
nickboldt: Should we mention "SDK, runtime, and debugger" here?

dependencies/che-devfile-registry/devfiles/05_dotnet/devfile.yaml
```diff
@@ -24,8 +30,8 @@ components:
         port: 5000
     mountSources: true
     volumes:
-      - name: dotnet
-        containerPath: "/home/user"
+      - name: nuget
+        containerPath: "/home/jboss/.nuget"
```
nickboldt: Why the .nuget subfolder? Also do we need to pre-fetch anything into that folder for offline/Airgap?

svor: .nuget is like .m2 for maven - it is used for dependencies and it is created automatically in user home directory. I wouldn't  pre-fetch something into that folder, it depends on what user needs.

dependencies/che-devfile-registry/devfiles/05_dotnet/meta.yaml
svor: I don't think we have to describe plugins here (C# language support, netcoredbg). If we have to write information about plugins than we need to review all descriptions for all devfiles. 
For example for java devfile we have `RHEL 8 Java stack with EAP 7.2, OpenJDK 1.8 and Maven 3.5` and nothing about debugger and jdt.ls

nickboldt: Fair point, but the debugger wasn't included in previous releases, so I was thinking it might be worth calling it out as a new feature.

svor: OK, the description was updated

PR #246, opened 2020-03-04
Add gradle volume; Add command to run endpoint

svor:
-  Adds a command to run endpoint into java11 plugins
-  Adds a volume for ~/.gradle folder into plugins where plugin-java11-rhel8:2.1 is used


svor:
@nickboldt take a look pls

nickboldt:
BTW according to @amisevsk in https://github.com/eclipse/che-jwtproxy/pull/20/files#r385727728

"Unsure if it's a plus or minus, but having a volume in the dockerfile means (some) OpenShift clusters will show an error event if the pod doesn't mount a corresponding volume in the podspec. If the volumes used are different between the two dockerfiles though, it'll make it hard to switch a deployment to use the rhel version."

svor:
@nickboldt That comment was about the volume in **Dockerfile** but this PR is about **Devfile**

nickboldt:
ah, good point. Thanks for setting me straight :D

PR #252, opened 2020-03-16
Change endpoint's name

svor:

### What does this PR do?
Changes endpoint's name from `5000/tcp` to `hello-endpoint`. 
Looks like the problem is because the name contains `/` symbol. 
screenshot-codeready-iokhrime-crw2 apps ocp44 codereadyqe com-2020 03 (1)
screenshot-codeready-iokhrime-crw2 apps ocp44 codereadyqe com-2020 03

The real problem should be fixed in upstream

https://issues.redhat.com/browse/CRW-717

cc @nickboldt


sunix:
This is a quick fix, we need a real fix in upstream.

svor:
@sunix thanks for the issue in upstream

> This is a quick fix, we need a real fix in upstream.

I've mentioned this in the description

sunix:
OK :) did a quick read of the quick fix :)

sunix:
upstream issue: https://github.com/eclipse/che/issues/16381

PR #253, opened 2020-03-18
Add commands for PHP Cake devfile

svor:
### What does this PR do?
Adds commands into PHP Cake devfile

https://issues.redhat.com/browse/CRW-723

cc @nickboldt

### DIFF
--- a/dependencies/che-devfile-registry/devfiles/05_php-cake/devfile.yaml
+++ b/dependencies/che-devfile-registry/devfiles/05_php-cake/devfile.yaml
@@ -37,22 +37,37 @@ commands:
   actions:
     - type: exec
       component: php
-      command: "composer install --no-interaction && wget https://get.symfony.com/cli/installer -O - | bash"
+      command: "composer install --no-interaction"
       workdir: "${CHE_PROJECTS_ROOT}/cakephp-ex"
-- 
-  name: Start Symfony Web Server
-  actions:
-    - type: exec
-      component: php
-      command: '$HOME/.symfony/bin/symfony server:start --no-tls'
-      workdir: '${CHE_PROJECTS_ROOT}/cakephp-ex'
 -
-  name: Stop Symfony Web Server
+  name: Start Apache Web Server
   actions:
     - type: exec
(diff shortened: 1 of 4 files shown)


dependencies/che-devfile-registry/devfiles/05_php-cake/devfile.yaml
```diff
@@ -24,8 +24,8 @@ components:
   memoryLimit: 512Mi
   mountSources: true
   endpoints:
-    - name: '8080/tcp'
-      port: 8080
+    - name: '8000-tcp'
```
nickboldt: haha good quickfix there :D

dependencies/che-devfile-registry/devfiles/05_php-cake/devfile.yaml
```diff
@@ -37,5 +37,39 @@ commands:
   actions:
     - type: exec
       component: php
-      command: "composer install 2>&1"
+      command: "composer install --no-interaction && wget https://get.symfony.com/cli/installer -O - | bash"
```
nickboldt: for airgap this might not work. Do we need to prefetch this and include it at build time?

svor: in airgap case - yes we do
Or we can replace commands to use Apache web server instead of Symfony.I think it is simpler way

svor:
@nickboldt PR is ready again

nickboldt:
Looks good but I bet `composer install --no-interaction` will fail in airgap unless we've already downloaded all the deps that the project needs

svor:
do you think that we need to download all deps that projects need, the same question for java projects where we have `mvn clean install` commands?

svor:
It is hard to collect all dependencies. I think in airgap user has to follow the documentation how to set up AirGap solution.

nickboldt:
Good point. User in airgap would have to ping their network administrator if `composer install` doesn't work.

Same as for maven projects that need to dl the internet of maven things.

PR #254, opened 2020-03-19
Replace '/' with '-' in endpoint's name

svor:

### What does this PR do?
Replaces '/' with '-' in endpoint's names in the devfile registry.

It is quick fix, the general fix should be done in https://github.com/eclipse/che/issues/16381.

https://issues.redhat.com/browse/CRW-717

cc @nickboldt


nickboldt:
thanks!

PR #255, opened 2020-03-19
Update the command to debug kitchensink-eap application

svor:

### What does this PR do?
Updates  **build and run in debug** command in **java-eap-maven** devfile to make debugger possible to attache to the remote process and debug eap application.

https://issues.redhat.com/browse/CRW-724

cc @nickboldt

### DIFF
--- a/dependencies/che-devfile-registry/devfiles/00_java-eap-maven/devfile.yaml
+++ b/dependencies/che-devfile-registry/devfiles/00_java-eap-maven/devfile.yaml
@@ -42,7 +42,11 @@ commands:
       -
         type: exec
         component: maven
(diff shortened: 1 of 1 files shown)


nickboldt:
other than line breaks (nice, much more legible) the change is to remove these flags:

```
 -b 0.0.0.0 --debug 8000
```

Is that correct?

svor:
@nickboldt because `-b 0.0.0.0 --debug 8000` doesn't work here, I don't know why it was added here
The main changes which was added by me is
`JAVA_OPTS_APPEND="-Xrunjdwp:transport=dt_socket,address=8000,server=y,suspend=n` 
It is not only line breaks :-)

PR #256, opened 2020-03-20
Patch version of the Dashboard

sleshchenko:
### What does this PR do?
Patches version of the Dashboard.
Upstream related changes were merged in https://github.com/eclipse/che/pull/16406

https://issues.redhat.com/browse/CRW-673

#### Is it tested? How?
Not fully. I mean I've done:
```bash
export JOB_NAME="sleshche-local"
export BUILD_NUMBER=1
export GIT_COMMIT=123456
export GIT_BRANCH=patchedVersion
mvn clean install
```
and checked that the build tomcat.zip contains patched version but I have not built and deploy CRW to check if Dashboard displays it
Screenshot_20200320_090628

### DIFF
--- a/assembly/codeready-workspaces-assembly-dashboard-war/pom.xml
+++ b/assembly/codeready-workspaces-assembly-dashboard-war/pom.xml
@@ -67,7 +67,7 @@
                 <version>1.8</version>
                 <executions>
                     <execution>
-                        <id>update-docs-version-product-json</id>
+                        <id>update-product-json</id>
                         <phase>validate</phase>
                         <goals>
                             <goal>run</goal>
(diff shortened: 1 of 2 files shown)


nickboldt:
thanks, this looks good. 

but it requires a new build of che-server 7.9.3, right?

dmytro-ndp:
We haven't released che-server:7.9.2. So, it could be included into it, IMHO.

nickboldt:
OK then we need a 7.9.2.RC2 build that includes https://github.com/eclipse/che/pull/16406 and anything else in 7.9.x branch. @vparfonov @mkuznyetsov can you respin 7.9.2?

nickboldt:
I can merge this if you think we need it but based on https://issues.redhat.com/browse/CRW-673?focusedCommentId=14005214&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14005214 seems like we don't need it. :)

sleshchenko:
>  seems like we don't need it. :)

I believe it's needed! The previous solution is not stable and kind of magic. You can rely on that on your own risk

PR #263, opened 2020-03-26
[Che 7.13] Updated Go stack's go.lintFlags to array

gautamkrishnar:
### What does this PR do?
Updates go stack's go.lintFlags parameter to use an array instead of string
Please see docs: https://github.com/microsoft/vscode-go#linter

While trying out Go stack on eclipse che hosted by Red Hat, user will recieve the folowing message:
Screenshot 2020-03-26 at 12 13 16 PM

#### Release Notes

N/A


nickboldt:
Looks like this depends on Che 7.13 so we will hold off merging this until we're on Che 7.13 (in a couple weeks)

PR #269, opened 2020-04-01
CRW-776 docker container for golang sets...

nickboldt:
CRW-776 docker container for golang sets GOPATH=/projects/.che/gopath so presumably we should do the same here since /go doesn't exist and isn't user-writeable

Change-Id: Ia0a681c77f666f47de4115da56f6292133ed410c


nickboldt:
dupe of #267

PR #274, opened 2020-04-06
Fail registry builds if a digest is not found

davidfestal:
### What does this PR do?

Fail registry builds if a digest is not found. To avoid produce a registry that in fact is silently containing image references without a digest. This would make this registry broken in regard to disconnected network compatibility.

Also updated mongodb image to the `1-57` minor version according to @nickboldt suggestion.


davidfestal:
Yeah, maybe this one `registry.access.redhat.com`  can be interesting.

PR #275, opened 2020-04-06
Fail registry builds if a digest is not...

nickboldt:
Fail registry builds if a digest is not found

Signed-off-by: David Festal <dfestal@redhat.com>

Try `registry.access.redhat.com` for rhscl images; fall back to registry-proxy.engineering.redhat.com/rh-osbs/ if could not resolve digest from public registries

Change-Id: I21e6ceb028eb76aa8d6d4b3fcda675cc38b1c5ee
Signed-off-by: David Festal <dfestal@redhat.com>


nickboldt:
enhanced version of #274

PR #281, opened 2020-04-23
CRW-1023 use smaller JDK 11 based Che server image 

nickboldt:
switch to ubi8/ubi-minimal base image then install JDK 11; makes the resulting image ~100M smaller

This is now more urgent as it's needed for https://issues.redhat.com/browse/CRW-1023

Change-Id: Ia5edd122bb4991d85e9a79abff5efc43f742307d


nickboldt:
Might be able to remove more stuff but I tried to run

```
rpm -qia|awk '$1=="Name" { n=$3} $1=="Size" {s=$3} $1=="Description" {print s  " " n }' |sort -n
```

and everything I tried to remove was a requirement of jre or systemd.

nickboldt:
Seems like this change is also required now that che server requires JDK 11

PR #283, opened 2020-04-30
Add vscode-yaml depedencies where needed

ericwill:
Currently the vscode-camelk and vscode-kubernetes-tools plugins
don't have the correct vscode-yaml dependency in their meta.yaml.
This commit adds them in to reflect the current state upstream.

Fixes CRW-835.


nickboldt:
Problem with this is that the above plugins are already live in 2.1.1 so ... we really need to bump the version here and release NEW plugins.

ericwill:
I'm not quite sure I follow...why do we need to release new plugins?

nickboldt:
if the existing plugin is already loaded, how will you force it to reload and fetch new extensions?

ericwill:
I understand now. Should we do it for all the versions of the Kubernetes and vscode-camelk plugins, or just the latest (1.1.0 and 0.0.12, respectively)?

nickboldt:
Yes for 1.1.0 -> 1.1.0a. Seems we missed that fix (though it's there in upstream)

https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools/1.1.0/meta.yaml

But for 0.0.12, I would instead pull newer plugins from che upstream https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/vscode-camelk/0.0.13/meta.yaml

and verify that we're using the latest vsix in both.

ericwill:
>     * https://github.com/redhat-developer/vscode-yaml/releases/download/0.7.2/redhat.vscode-yaml-0.7.2.vsix

Will do, but we have a download.jboss.org link for that one. :)

nickboldt:
download.jboss.org is better because it's more likely to have co-located sources in the same folder.

PR #286, opened 2020-05-20
streamline error handling in write_image_digest

kcrane:
### What does this PR do?

streamline error handling in write_image_digest

### DIFF
--- a/dependencies/che-devfile-registry/build/scripts/write_image_digests.sh
+++ b/dependencies/che-devfile-registry/build/scripts/write_image_digests.sh
@@ -17,8 +17,12 @@ LOG_FILE="$(mktemp)" && trap "rm -f $LOG_FILE" EXIT
 function handle_error() {
   local yaml_file="$1"
   local image_url="$2"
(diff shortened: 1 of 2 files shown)


nickboldt:
Do we need this in the equivalent file for the plugin registry too?

dependencies/che-devfile-registry/build/scripts/write_image_digests.sh
```diff
@@ -17,15 +17,9 @@ LOG_FILE="$(mktemp)" && trap "rm -f $LOG_FILE" EXIT
 function handle_error() {
   local yaml_file="$1"
   local image_url="$2"
-  if [[ -z "$(tail -1 $LOG_FILE | grep -v "no image found in manifest list for architecture $ARCH")" ]] ; then
+  if [[ ! -z "$($SCRIPT_DIR/find_image.sh "${image_url%:*}" x86_64 | jq -r '.Digest')" ]] ; then
```
nickboldt, line 20: Warning message is no longer correct... If we check for x86_64 and then say "not found for $ARCH", that's misleading.

kcrane: updated with more robust logging.

PR #290, opened 2020-05-22
CRW-871 remove refs to stack-analysis.sh; switch to using OSBS builds of EAP 7.3.1/XP1.0; for Fuse, use JDK 8; for the rest switch to JDK 11

nickboldt:
CRW-871 remove refs to stack-analysis.sh; switch to using OSBS builds of EAP 7.3.1/XP1.0; for Fuse, use JDK 8; for the rest switch to JDK 11

Change-Id: I498cea8df63b3591a4f7837e5f4fb1664f25cdb3

### DIFF
--- a/dependencies/che-devfile-registry/devfiles/00_java-eap-maven/devfile.yaml
+++ b/dependencies/che-devfile-registry/devfiles/00_java-eap-maven/devfile.yaml
@@ -11,11 +11,17 @@ projects:
 components:
   -
     type: chePlugin
-    id: redhat/java8/latest
+    id: redhat/java11/latest
+  -
+    # NOTE: instead of the old stack-analysis script, should be able to use the latest dependency-analysis plugin instead
+    type: chePlugin
+    id: redhat/dependency-analytics/latest
   -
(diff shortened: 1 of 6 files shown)


nickboldt:
I have not checked if:
* 5 java projects still run after switching to JDK 11
* fuse commands work using `scl enable rh-maven35 '...'` wrapper 
* stack analysis plugin works as expected without having to point **explicitly** at the pom & folder to analyse

svor:
@nickboldt should **meta.yaml** be updated for each devfile from these changes

nickboldt:
Yes, we should remember to update meta.yaml if these changes are accepted.

As to `redhat/java8` yes, or else we could use `plugin-java11-rhel8` if we don't need a JDK8 runtime for JDTLS & debugging. Is JDK 11 backward compatible w/ JDK8 in terms of the java8 plugin?

If not... maybe we use `redhat-openjdk-18/openjdk18-openshift:1.8` from https://catalog.redhat.com/software/containers/detail/58ada5701fbe981673cd6b10

PR #296, opened 2020-05-29
swap openjdk images for openj9 on p and z

kcrane:
### What does this PR do?

swap openjdk images for openj9 on p and z

### DIFF
--- a/dependencies/che-devfile-registry/build/scripts/swap_images.sh
+++ b/dependencies/che-devfile-registry/build/scripts/swap_images.sh
@@ -9,11 +9,11 @@
 SCRIPT_DIR=$(cd "$(dirname "$0")" || exit; pwd)
 YAML_ROOT="$1"
 
-if [[ "$(uname -m)" != "s390x" ]] && [[ "$(uname -m)" != "ppc64le" ]] ; then
+if [[ "$(uname -m)" == "x86_64" ]] ; then
   exit 0
 fi
 
 readarray -d '' devfiles < <($SCRIPT_DIR/list_yaml.sh "$YAML_ROOT" | tr '\n' '\0')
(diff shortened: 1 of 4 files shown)


dependencies/che-devfile-registry/build/scripts/swap_images.sh
```diff
@@ -0,0 +1,19 @@
...
+YAML_ROOT="$1"
+
+if [[ "$(uname -m)" != "s390x" ]] && [[ "$(uname -m)" != "ppc64le" ]] ; then
+  exit 0
+fi
+
+readarray -d '' devfiles < <($SCRIPT_DIR/list_yaml.sh "$YAML_ROOT" | tr '\n' '\0')
+sed -E -i 's/(.*image:.*)jboss-eap-7-eap-xp1-openjdk11-openshift-rhel8:latest(.*)/\1jboss-eap-7-eap73-openj9-11-openshift-rhel8:7.3.0-2\2/g' ${devfiles[@]}
+sed -E -i 's/(.*image:.*)jboss-eap-7-eap-xp1-openjdk8-openshift-rhel7:1.0(.*)/\1jboss-eap-7-eap73-openj9-11-openshift-rhel8:7.3.0-2\2/g' ${devfiles[@]}
+sed -E -i 's/(.*value:.*)sso74-openshift-rhel8:7.4(.*)/\1sso74-openj9-openshift-rhel8:7.4\2/g' ${devfiles[@]}
```
nickboldt, line 19: Not a fan of hardcoding the image tags here... maybe we should use `7.3.0` and use getLatestImages.sh to find the actual tag value (currently `7.3.0-2` but might be different on Z vs. P).

Also wondering what we do when the x image exists on RHCC but the Z or P image is still only on OSBS. I suspect this change might result in pointers to non-existent images.

kcrane: yup, agree on both points... will update.

PR #321, opened 2020-07-10
🚧 Don't hard-code docs version related to converting CRW 1.2 workspace 2.0

sleshchenko:
### What does this PR do?
I occasionally figured out that we have such a link which is hard-coded to a specific CRW version which I suppose is the wrong thing to do since recently we have updated the corresponding article in upstream https://github.com/eclipse/che-docs/pull/1358 which is going to appear in CRW 2.2 Docs.

With this PR I just wanted to raise that outdated documentation referencing and it's just suggestions I'm not sure about like if it's the right link we'll have working in the end - @rkratky @nickboldt feel free to update the PR

https://github.com/eclipse/che/issues/17218

### DIFF
--- a/assembly/branding/product.json.template
+++ b/assembly/branding/product.json.template
@@ -32,7 +32,7 @@
     "workspace" : "@@crw.docs.baseurl@@/html-single/end-user_guide/index#workspaces-overview_crw",
     "factory" : "@@crw.docs.baseurl@@/html-single/end-user_guide/index#overriding-devfile-values-using-factory-parameters_configuring-a-workspace-using-a-devfile",
     "organization": "@@crw.docs.baseurl@@/html-single/administration_guide/index#managing-users_crw",
-    "converting": "https://access.redhat.com/documentation/en-us/red_hat_codeready_workspaces/2.0/html-single/end-user_guide/index#converting-a-codeready-workspaces-1.2-workspace-to-a-codeready-workspaces-2.0-devfile",
+    "converting": "@@crw.docs.baseurl@@/html-single/end-user_guide/index#converting-a-codeready-workspaces-1.2-workspace-to-a-codeready-workspaces-2.0-devfile",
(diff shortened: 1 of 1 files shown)


dmytro-ndp:
@sleshchenko: are you sure about the path to the doc `end-user_guide/index#converting-a-codeready-workspaces-1.2-workspace-to-a-codeready-workspaces-2.0-devfile`?

sleshchenko:
@dmytro-ndp Thanks for raising it again.
If you already may saw in PR description - I'm not sure

> With this PR I just wanted to raise that outdated documentation referencing and it's just suggestions I'm not sure about like if it's the right link we'll have working in the end

sleshchenko:
I've checked and I haven't found that article in Doc for CRW 2.2
But if it's removed - should Dashboard provide user's reference to old doc that has outdated description?
Or we remove that functionality from Dashboard and upstream docs and expect that every instance already does not have that deprecated workspaces?

nickboldt:
See also https://issues.redhat.com/browse/RHDEVDOCS-1969 ... there's no need to have doc about how to convert from 1.2 to 2.2, since the correct path is:

* 1.2 -> 2.0
* 2.0 -> 2.1
* 2.1 -> 2.2

That's why I intentionally hardcoded the link to refer to the 2.0 docs.

sleshchenko:
But there is no guarantee that every user migrated their workspace before the next CRW version upgrade, right?

@nickboldt So, what your opinion about that docs article and functionality in Dashboard in upstream? Should it be removed at this point?

sleshchenko:
That PR is not going to be merged despite of future for that in upstream

nickboldt:
I might mark it deprecated in the current sprint and remove it in 3 sprints' time. That's the Che Way. :)

1.2 is past EOL so for 2.4 it would make sense to no longer link to the 2.0 docs to tell people how to move off Che 6.x based workspaces. 

But just as we still support OCP 3.11 while 4.5 is out, so too we have to at least expect that not everyone can immediately move to the bleeding edge immediately.

PR #324, opened 2020-07-16
Adding build-time verification of che-machine-exec and che-theia exist in plugin registry

dipika21:
Hi @nickboldt  : Based on my understanding on this Jenkinsfile I have added code at line number 237, rather than creating another variable I used existing variable ${VER_CHE}. 

### What does this PR do?
Modified Jenkinsfile for build-time verification of che-machine-exec and che-theia exist in plugin registry.

In maven process, if specified che.version is not in registry then it will add che version fo che-machine-exec and che-theia by running add_che_plugin.sh file.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -234,10 +234,25 @@ timeout(240) {
 		// unpack asset-*.tgz into folder where mvn can access it
 		// use that content when building assembly main and ws assembly?
 
+		git checkout --track origin/''' + branchToBuildCRW + ''' || true
+		export GITHUB_TOKEN=''' + GITHUB_TOKEN + ''' # echo "''' + GITHUB_TOKEN + '''"
+		git config user.email "nickboldt+devstudio-release@gmail.com"
+		git config user.name "Red Hat Devstudio Release Bot"
+		git config --global push.default matching
+
+		// SOLVED :: Fatal: Could not read Username for "https://github.com", No such device or address :: https://github.com/github/hub/issues/1644
+		git remote -v
+		git config --global hub.protocol https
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -234,6 +234,11 @@ timeout(240) {
 		// unpack asset-*.tgz into folder where mvn can access it
 		// use that content when building assembly main and ws assembly?
 
+		// Check if che-machine-exec and che-theia plugins are current in upstream repo and if not, add them
+		if [[ ! -d "dependencies/che-plugin-registry/v3/plugins/eclipse/che-machine-exec-plugin/${VER_CHE}" ]] || [[ ! -d "dependencies/che-plugin-registry/v3/plugins/eclipse/che-theia/${VER_CHE}" ]]; then
+			./dependencies/che-plugin-registry/build/scripts/add_che_plugins.sh ${VER_CHE}
```
nickboldt, line 239: script can't be run from this folder. 

Better to pushd into the `./dependencies/che-plugin-registry` folder and then run it from there.

nickboldt: also it occurs to me that you might not have also I doubt this will be able to successfully commit the change back to the repo.

You need to configure the ${WORKSPACE}/${CRW_path} folder like this (all 10 lines):

https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-plugin-registry/Jenkinsfile#L81-L91

nickboldt:
thanks, looks good now.

nickboldt:
Note, needed a few more changes to make this actually work.

* https://github.com/redhat-developer/codeready-workspaces/commit/757929f90efb035acacfd683956706b2b363919b
* https://github.com/redhat-developer/codeready-workspaces/commit/ff8a6fbfa17e6fb82e15c9b21ce59a064ad5e9c8

PR #325, opened 2020-07-22
CRW-1093 refactor Dockerfile and...

nickboldt:
CRW-1093 refactor Dockerfile and entrypoint.sh to align to upstream changes

Change-Id: I89fba06432563a8a81ad95e88452fad9773d5cd7

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,30 +15,34 @@ USER root
 RUN microdnf install java-11-openjdk-headless tar gzip shadow-utils findutils && \
     microdnf update -y gnutls && \
     microdnf -y clean all && rm -rf /var/cache/yum && echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages" && \
-    adduser -G root jboss && mkdir -p /home/jboss/codeready
+    adduser -G root user && mkdir -p /home/user/codeready && \
+    # fix certs & dir permissions
+    cp /etc/pki/ca-trust/extracted/java/cacerts /home/user/cacerts && chmod 644 /home/user/cacerts
+ENV CHE_HOME=/home/user/codeready
 ENV JAVA_HOME=/usr/lib/jvm/jre
+ENV JAVA_TRUST_STORE=/etc/pki/ca-trust/extracted/java/cacerts
 
 COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
 
 # NOTE: if built in Brew, use get-sources-jenkins.sh to pull latest
 # OR, if you intend to build the Che Server tarball locally, 
 # see https://github.com/redhat-developer/codeready-workspaces-productization/blob/master/devdoc/building/building-crw.adoc#make-changes-to-crw-and-re-deploy-to-minishift
 # then copy /home/${USER}/projects/codeready-workspaces/assembly/codeready-workspaces-assembly-main/target/codeready-workspaces-assembly-main.tar.gz into this folder
 COPY assembly/codeready-workspaces-assembly-main/target/codeready-workspaces-assembly-main.tar.gz /tmp/codeready-workspaces-assembly-main.tar.gz
-RUN tar xzf /tmp/codeready-workspaces-assembly-main.tar.gz --transform="s#.*codeready-workspaces-assembly-main/*##" -C /home/jboss/codeready && \
+RUN tar xzf /tmp/codeready-workspaces-assembly-main.tar.gz --transform="s#.*codeready-workspaces-assembly-main/*##" -C /home/user/codeready && \
     rm -f /tmp/codeready-workspaces-assembly-main.tar.gz
-# this should fail if the startup script is not found in correct path /home/jboss/codeready/tomcat/bin/catalina.sh
-RUN java -version && \
-    echo -n "Server startup script in: " && find /home/jboss/codeready -name catalina.sh | grep -z /home/jboss/codeready/tomcat/bin/catalina.sh && \
-    # fix certs & dir permissions
-    cp /etc/pki/ca-trust/extracted/java/cacerts /home/jboss/cacerts && chmod 644 /home/jboss/cacerts && \
-    mkdir -p /logs /data && \
-    chgrp -R 0     /home/jboss /data /logs && \
-    chmod -R g+rwX /home/jboss /data /logs && \
-    chown -R jboss /home/jboss
-
-USER jboss
-ENTRYPOINT ["/entrypoint.sh"]
+# this should fail if the startup script is not found in correct path /home/user/codeready/tomcat/bin/catalina.sh
+RUN mkdir /logs /data && \
+    chmod 0777 /logs /data && \
+    chgrp -R 0 /home/user /logs /data && \
+    chown -R user /home/user && \
+    chmod -R g+rwX /home/user && \
+    find /home/user -type d -exec chmod 777 {} \; && \
+    java -version && echo -n "Server startup script in: " && \
+    find /home/user/codeready -name catalina.sh | grep -z /home/user/codeready/tomcat/bin/catalina.sh
+    
+USER user
 
 ENV SUMMARY="Red Hat CodeReady Workspaces Server container" \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -15,30 +15,34 @@ USER root
 RUN microdnf install java-11-openjdk-headless tar gzip shadow-utils findutils && \
     microdnf update -y gnutls && \
     microdnf -y clean all && rm -rf /var/cache/yum && echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages" && \
-    adduser -G root jboss && mkdir -p /home/jboss/codeready
+    adduser -G root user && mkdir -p /home/user/codeready && \
+    # fix certs & dir permissions
+    cp /etc/pki/ca-trust/extracted/java/cacerts /home/user/cacerts && chmod 644 /home/user/cacerts
+ENV CHE_HOME=/home/user/codeready
```
nickboldt, line 21: new overrides for default che entrypoint.sh script:
* CHE_HOME
* JAVA_TRUST_STORE

Dockerfile
```diff
@@ -15,30 +15,34 @@ USER root
 RUN microdnf install java-11-openjdk-headless tar gzip shadow-utils findutils && \
     microdnf update -y gnutls && \
     microdnf -y clean all && rm -rf /var/cache/yum && echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages" && \
-    adduser -G root jboss && mkdir -p /home/jboss/codeready
+    adduser -G root user && mkdir -p /home/user/codeready && \
```
nickboldt, line 18: switch to user `user` so we're consistent w/ upstream

Dockerfile
```diff
@@ -15,30 +15,34 @@ USER root
...
 # then copy /home/${USER}/projects/codeready-workspaces/assembly/codeready-workspaces-assembly-main/target/codeready-workspaces-assembly-main.tar.gz into this folder
 COPY assembly/codeready-workspaces-assembly-main/target/codeready-workspaces-assembly-main.tar.gz /tmp/codeready-workspaces-assembly-main.tar.gz
-RUN tar xzf /tmp/codeready-workspaces-assembly-main.tar.gz --transform="s#.*codeready-workspaces-assembly-main/*##" -C /home/jboss/codeready && \
+RUN tar xzf /tmp/codeready-workspaces-assembly-main.tar.gz --transform="s#.*codeready-workspaces-assembly-main/*##" -C /home/user/codeready && \
     rm -f /tmp/codeready-workspaces-assembly-main.tar.gz
-# this should fail if the startup script is not found in correct path /home/jboss/codeready/tomcat/bin/catalina.sh
-RUN java -version && \
-    echo -n "Server startup script in: " && find /home/jboss/codeready -name catalina.sh | grep -z /home/jboss/codeready/tomcat/bin/catalina.sh && \
-    # fix certs & dir permissions
-    cp /etc/pki/ca-trust/extracted/java/cacerts /home/jboss/cacerts && chmod 644 /home/jboss/cacerts && \
```
nickboldt, line 34: moved up above... but do we still need this? @skabashnyuk

entrypoint.sh
```diff
@@ -229,11 +229,11 @@ init() {
 
   if [ -z "$CHE_HOME" ]; then
```
nickboldt, line 230: for CRW we set this in Dockerfile so that we can override the old defaults.

entrypoint.sh
```diff
@@ -291,28 +291,50 @@ init() {
 }
 
 add_cert_to_truststore() {
-  if [ "${CHE_SELF__SIGNED__CERT}" != "" ]; then
-    DEFAULT_JAVA_TRUST_STORE=/etc/pki/ca-trust/extracted/java/cacerts
-    DEFAULT_JAVA_TRUST_STOREPASS="changeit"
+  DEFAULT_JAVA_TRUST_STORE=${JAVA_TRUST_STORE:-${JAVA_HOME}/lib/security/cacerts}
```
nickboldt, line 294: override default in CRW Dockerfile since that path doesn't work OOTB in ubi8-minimal

entrypoint.sh
```diff
@@ -291,28 +291,50 @@ init() {
...
+  fi
 
-    echo "$CHE_SELF__SIGNED__CERT" > $SELF_SIGNED_CERT
-
-    echo yes | keytool -keystore $JAVA_TRUST_STORE -importcert -alias HOSTDOMAIN -file $SELF_SIGNED_CERT -storepass $DEFAULT_JAVA_TRUST_STOREPASS > /dev/null
+  echo "$1" > $SELF_SIGNED_CERT
 
-    # DOES NOT WORK IN OPENSHIFT - maybe we need to add anyuid fix? 
-    # allow only read by all groups
-    # if [[ -w $JAVA_TRUST_STORE ]]; then chmod 444 $JAVA_TRUST_STORE; fi
```
nickboldt, line 310: unclear if this commented block is outdated since the next line suggests this DOES work

nickboldt:
See also https://github.com/eclipse/che/pull/17465

Dockerfile
```diff
@@ -12,33 +12,38 @@
...
+ENV JAVA_HOME=/usr/lib/jvm/jre
 RUN microdnf install java-11-openjdk-headless tar gzip shadow-utils findutils && \
     microdnf update -y gnutls && \
     microdnf -y clean all && rm -rf /var/cache/yum && echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages" && \
-    adduser -G root jboss && mkdir -p /home/jboss/codeready
-ENV JAVA_HOME=/usr/lib/jvm/jre
+    adduser -G root user && mkdir -p /home/user/codeready && \
+    # fix certs & dir permissions - see file references in entrypoint.sh
+    for d in /home/user ${JAVA_HOME}/lib/security; do \
+      mkdir -p ${d}; cp /etc/pki/ca-trust/extracted/java/cacerts ${d}/cacerts && chmod 644 ${d}/cacerts; \
```
nickboldt, line 23: rather than setting an override for the path to default cacerts file (in case it doesn't exist) just create it.

entrypoint.sh
skabashnyuk: is this a copy of code from upstream?

l0rd: That's what I was wondering as well. Are you making the entrypoints (upstream/downstream) identical? Because thy should be identical. And if they are identical why are we duplicating the code in 2 repositories?

nickboldt: > is this a copy of code from upstream?

It probably was at some point but because we have a COPY of the file it's gotten outdated. eventually we'd have this synced from upstream directly, but first I want to fix it and verify it works before I set up new sync rules for pulling files from upstream directly.

As there's no rhel.Dockerfile in https://github.com/eclipse/che/tree/master/dockerfiles/che it's harder to keep that and entrypoint.sh current w.r.t. downstream. 

Would you be open to having a rhel.Dockefile in the above folder so that I can simply sync that file + entrypoint.sh to our midstream CRW GH repo?

nickboldt: > That's what I was wondering as well. Are you making the entrypoints (upstream/downstream) identical? Because thy should be identical. And if they are identical why are we duplicating the code in 2 repositories?

Yes, if you look at the two PRs linked here you'll see the output is to have both the upstream alpine based Dockerfile and the downstream RHEL based dockerfile using the same entrypoint.sh script.

skabashnyuk:
@nickboldt wdyt about moving rhel.Devfile to upstream?

l0rd:
@nickboldt @skabashnyuk I am definately +1 to move `rhel.Devfile` upstream

nickboldt:
going to assume you meant `rhel.Dockerfile` and will throw a PR your way for upstream.

l0rd:
> going to assume you meant `rhel.Dockerfile` and will throw a PR your way for upstream."

yes sorry that was about the Dockerfile and not the devfile...

nickboldt:
Once the above is approved & merged (Happy path failing again) https://github.com/eclipse/che/pull/17494 we can merge this and verify it's working.

PR #341, opened 2020-09-15
latest golang plugin should be using latest stacks-golang

kcrane:
### What does this PR do?

latest golang plugin should be using latest stacks-golang

fix error:
failed to fetch plugin meta.yaml from URL 'https://plugin-registry-crw.apps.qe2-s390x.psi.redhat.com/v3/plugins/ms-vscode/go/latest/meta.yaml'


nickboldt:
:+1:

PR #345, opened 2020-09-18
fix spring boot devfile to have generateName

kcrane:
### What does this PR do?

fix springboot devfile to have generateName (otherwise won't open).  Verified this is a one-off (all others have generateName).


nickboldt:
doublechecked and yes, all the other 18 devfiles already have `.metadata.generateName` instead of `.metadata.name`.

PR #353, opened 2020-10-12
Remove thorntail stack

svor:

### What does this PR do?
Remove Thorntail stack as  it is no longer supported. 

https://issues.redhat.com/browse/CRW-1242


nickboldt:
@ericwill this won't actually solve CRW-1242, but we can remove thorntail if you agree we no longer want to provide it. See https://github.com/eclipse/che/issues/17961 -- is thorntail no longer supported?

nickboldt (commented):
+0, this doesn't solve the JIRA but might be worth doing if we want to drop thorntail anyway.

nickboldt:
@ericwill are we dropping thorntail in che plugin reg upstream?

PR #358, opened 2020-10-20
Upgrade vscode-xml to 0.13.0

ericwill:
Fixes https://issues.redhat.com/browse/CRW-1305


apupier (approved):
sounds good by reading code but I do not have a CodeReady Workspaces development environment so cannot test it.

nickboldt:
Don't you love that we only include a single version of plugin now, so it's easier to update two files w/o having to mark older ones deprecated? Approved.

PR #373, opened 2020-11-19
Introduce ES Lint 2.1.1

vitaliy-guliy:
Signed-off-by: Vitaliy Gulyy <vgulyy@redhat.com>

### What does this PR do?
Adds ES Lint 2.1.1
Copy of the plugin in Che Plugin Registry https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins/dbaeumer/vscode-eslint

The only difference is that the plugin icon is loaded from Git Hub instead of getting it from Open VSX registry.

https://issues.redhat.com/browse/CRW-1357


mkuznyetsov (approved):

dependencies/che-plugin-registry/v3/plugins/dbaeumer/vscode-eslint/2.1.1/meta.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+publisher: dbaeumer
+name: vscode-eslint
+version: 2.1.1
+type: VS Code extension
+displayName: ESLint
+title: ESLint JavaScript
+description: Integrates ESLint JavaScript into Che-Theia.
+icon: https://raw.githubusercontent.com/microsoft/vscode-eslint/1e15d3495da89072d48cf583d48d92829f0c4b82/eslint_icon.png
```
nickboldt, line 9: This is the wrong icon... in CRW we suppress all the rando icons in the plugins view. 

Or is it time to revisit that decision? The original reason was that most of the plugins were all using the same Che icon (or CRW icon) and it looked busy/crowded/bad UX. 

But I see in Che 7.21 this is no longer the case... so I've opened https://issues.redhat.com/browse/CRW-1385 to discuss a more complete solution for 2.6 so we don't have a hodgepodge of some with icons and some without.

vitaliy-guliy: Now we are able to let the plugin to display his icon. I am +1 on the improvement if there was only one reason related to similar icons that you described.

nickboldt: I wonder if we should COPY the plugin reg images into this repo so they're all co-located in the same repo, rather than having links to 3rd party binary assets.

vitaliy-guliy: I think it's possible to copy the images (if license permits), but we need also to improve Che-Theia a bit. As we haven't tried to load the icons from the plugin registry, except the default Che icon.

nickboldt: See https://issues.redhat.com/browse/CRW-1385 :) seems that as of a recent Che build, you *can* load unique icons for thhe plugins.

PR #394, opened 2020-12-28
Use a dedicated Che Dashboard Version

sleshchenko:
### What does this PR do?
Eclipse Che has a plan to switch to a new React-based Dashboard in 7.25.
But the decision that CRW 2.7(Che 7.26) should switch to a new dashboard is not made.
So, this PR introduces a dedicated parameter to be able to stick to an old Dashboard version which is angular-based.

P.S. Probably more changes are needed to make it work, like setting this parameter somewhere...
This PR is more to track all stuff we need to track and after the decision is made - it will need to be closed or finalized.

It's related to switch to Dashboard Next within https://github.com/eclipse/che/issues/18348

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,6 +10,7 @@ import groovy.transform.Field
 @Field String branchToBuildDev = "refs/tags/19"
 @Field String branchToBuildParent = "refs/tags/7.15.0"
 @Field String branchToBuildChe = "7.20.x"
+@Field String branchToBuildCheDashboard = "7.20.x"
 @Field String MIDSTM_BRANCH = "crw-2.5-rhel-8" // target branch in GH repo, eg., crw-2.5-rhel-8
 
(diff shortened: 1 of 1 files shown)


sleshchenko:
@nickboldt Could you take care of making CRW depending on the old dashboard version.
I tried to help but not sure if my changes makes sense, Mykhailo notified me that CRW Jenkins is moving. So, we're going to replace dashboard in upstream at the beginning of 7.26-snapshot. So, CRW should depend on 7.25.x or 7.24.x <- The last actual change is made in 7.24.x https://github.com/eclipse/che-dashboard/commits/master

If any info/help is needed - let me know

nickboldt:
replaced by https://gitlab.cee.redhat.com/codeready-workspaces/crw-jenkins/-/merge_requests/166 (jenkinsfile has moved)

PR #402, opened 2021-01-13
Add icons into the plugin registry

ericwill:

### What does this PR do?
Adds icons locally and reference them in meta.yamls

CRW-1385


nickboldt:
Assuming that all the new images are correct and the new references are correct, then +1.

PR #405, opened 2021-01-15
Align Dockerfile with upstream to prevent invalid images from being built

ericwill:

### What does this PR do?

Prevents errors like:

`standard_init_linux.go:211: exec user process caused "no such file or directory"`

from occurring when building with the alpine based Dockerfile.

N/A


benoitf (approved):
PR is perfectly fine

But there is a problem of downstreaming, alpine Dockerfile shouldn't be kept downstream

cc @nickboldt

nickboldt:
Merged, but I agree w/ Florent -- we shouldn't have these files in downstream. See https://issues.redhat.com/browse/CRW-1498 for followup work

PR #416, opened 2021-01-22
Add skip_if_unavailable=True for content sets behind the VPN

ericwill:

### What does this PR do?
Allows non-RH people to build the rhel.Dockerfile without being behind the Red Hat VPN.

Part of RHDEVDOCS-2513


nickboldt:
not needed for 2.6, but harmless. so merging.

PR #430, opened 2021-02-18
For power : bumping up memory in plugin registry 

ghatwala:
For https://issues.redhat.com/browse/CRW-1475 bumping up theia memory in plugin registry for power

1. As per above JIRA , we needed more memory in plugin registry image for "che-plugin-registry/v3/plugins/eclipse/che-theia/next/meta.yaml"
2. Invoke new script bump_memory.sh for power in rhel.dockerfile
Signed-off-by: Amit Ghatwal <ghatwala@us.ibm.com>

### DIFF
--- a/dependencies/che-plugin-registry/build/scripts/bump_memory.sh
+++ b/dependencies/che-plugin-registry/build/scripts/bump_memory.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
+#
+# Copyright (c) 2018-2021 Red Hat, Inc.
+# This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


dependencies/che-plugin-registry/build/scripts/bump_memory.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# available under the terms of the Eclipse Public License 2.0
```
nickboldt, line 2: please add copyright date and holder

ghatwala: added copyright header - @nickboldt

PR #461, opened 2021-04-01
Add resource limits/requests to the plugins

svor:
### What does this PR do?
Adds `cpuLimit` and `cpuRequest` into plugin yamls. All values were ported from the upstream.

https://issues.redhat.com/browse/CRW-1637

### DIFF
--- a/dependencies/che-plugin-registry/v3/plugins/che-incubator/cpptools/0.1.0/meta.yaml
+++ b/dependencies/che-plugin-registry/v3/plugins/che-incubator/cpptools/0.1.0/meta.yaml
@@ -14,7 +14,9 @@ spec:
   containers:
     - image: "registry.redhat.io/codeready-workspaces/stacks-cpp-rhel8:2.9"
       name: cpp-plugins
-      memoryLimit: '512Mi'
+      memoryLimit: 512Mi
+      cpuLimit: 500m
+      cpuRequest: 30m
       args:
         - sh
         - -c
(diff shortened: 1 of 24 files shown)


ghatwala:
@svor  - generic query about this - how did you find/calculate/deduce on the values mentioned in all plugins here ? Am guessing unless we try these on all arches , we wouldn't know , correct ? : ) 
```
cpuLimit: xxm
cpuRequest: xxm
```

svor:
> @svor - generic query about this - how did you find/calculate/deduce on the values mentioned in all plugins here ? Am guessing unless we try these on all arches , we wouldn't know , correct ? : )

@ghatwala this investigation was done in the upstream https://github.com/eclipse-che/che-plugin-registry/pull/774. And you are right we should tests these changes on all arches to guarantee that it doesn't break anything.

ghatwala (commented):
+1 , for trying these on all arches, to check if it breaks anywhere .

nickboldt:
I'm +1 to merge this but was just wondering what the units are and what these actually do...

```
cpuLimit: 500m
cpuRequest: 30m
```

Does that mean this plugin will ask for a minimum of 0.030 and maximum of 0.500 virtual CPUs? Is the "m" for milli, as in 1/1000th of a virtual CPU?

ericwill:
> I'm +1 to merge this but was just wondering what the units are and what these actually do...
> 
> ```
> cpuLimit: 500m
> cpuRequest: 30m
> ```
> 
> Does that mean this plugin will ask for a minimum of 0.030 and maximum of 0.500 virtual CPUs? Is the "m" for milli, as in 1/1000th of a virtual CPU?

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu

Basically the `m` stands for millicores, so 100m is 0.1 of a CPU core/hyperthread

svor:
@nickboldt 
cpuLimit -> to make sure that container never go above the value
cpuRequest -> is what guaranteed by k8s

nickboldt:
cherrypicked to 2.8 branch as 28fd926e

PR #466, opened 2021-04-02
fix(index.json): Fix devfile ordering in dev environment

sunix:
### What does this PR do?
Fix devfile ordering in dev environment:
- local build with podman
- build in a dogfooding Che workspace

Selection_313


sunix:
if it works for you I can also apply the same in upstream

nickboldt:
Thanks, tested locally and works as expected. I'll +1 the same fix in upstream if you link me to that PR. :)

PR #476, opened 2021-04-26
Deprecate the asciidoctor-vscode plugin

ericwill:

### What does this PR do?
Deprecate this plugin as it doesn't seem to have a good reason for being there, and is broken.

Fixes [CRW-1701](https://issues.redhat.com/browse/CRW-1701)


nickboldt:
Seems legit, if there's no real reason to have this in there.

nickboldt (approved):

PR #481, opened 2021-04-28
Add in seperate dashboard image into product scripts

JPinkney:
Signed-off-by: Josh Pinkney <joshpinkney@gmail.com>

### What does this PR do?
This PR includes the dashboard image in the product scripts


product/getLatestImageTags.sh
```diff
@@ -65,6 +65,7 @@ CRW_CONTAINERS_RHEC="\
 codeready-workspaces/configbump-rhel8 \
 codeready-workspaces/crw-2-rhel8-operator \
 codeready-workspaces/crw-2-rhel8-operator-metadata \
+codeready-workspaces/dashboard-rhel8 \
```
nickboldt, line 68: same change needed on line 105. should really refactor so that the two variants for operator + metadata are transformative rules rather than a huge string copied out 2x, but at minimum need the dashboard listed in both places.

nickboldt:
once the above tweak is done, +1. prefer to not merge this until we have a dashboard build in brew & quay, so as to avoid breakages and also verify this change works as expected.

PR #482, opened 2021-04-30
Convert midstream repo to use upstream plugin registry automation

ericwill:
### What does this PR do?
Converts the CRW plugin registry to use the upstream TypeScript based plugin registry automation.

[CRW-1648](https://issues.redhat.com/browse/CRW-1648)


benoitf:
I'm wondering here if we should publish for example upstream library to npmjs and just use it downstream

So we only keep downstream 3 files like che-*.yaml files and we just invoke the generator from these files
And sync is automatically made (only need to bump library version)

ericwill:
> I'm wondering here if we should publish for example upstream library to npmjs and just use it downstream
> 
> So we only keep downstream 3 files like che-*.yaml files and we just invoke the generator from these files
> And sync is automatically made (only need to bump library version)

I think it's a good idea and worth doing, but for 2.9 let's just do a manual sync to get things in place. We can improve after that :+1:

benoitf:
ok, I was just thinking it's much more work to do manual sync than just publishing/executing the tool ?

for example in this current PR we're still using `extensions:` and there is no dependency as in upstream now so it will require a new copy of .ts files, etc ?

ericwill:
> ok, I was just thinking it's much more work to do manual sync than just publishing/executing the tool ?
> 
> for example in this current PR we're still using `extensions:` and there is no dependency as in upstream now so it will require a new copy of .ts files, etc ?

Well what takes the longest time is the `che-theia-plugins.yaml` as it's not a 1:1 subset of upstream, but that has to be done for both cases anyway. I've already started and should be done by this afternoon. Everything else is just copy/paste with a few small changes.

How would it look if we use the upstream tool as a published npm package? Will we still be able to build in a disconnected environment (brew)?

benoitf:
>How would it look if we use the upstream tool as a published npm package? Will we still be able to build in a disconnected environment (brew)?

if we publish to npm registry, we will still add that dependency in yarn v2 dependencies of this downstream repository so it'll be there like any other javascript dependency

benoitf:
> Well what takes the longest time is the che-theia-plugins.yaml as it's not a 1:1 subset of upstream, but that has to be done for both cases anyway. I've already started and should be done by this afternoon. Everything else is just copy/paste with a few small changes.

here I think it could be done automatically (improvement to made) to just have stuff automatically updated from the upstream yaml file

ericwill:
> > How would it look if we use the upstream tool as a published npm package? Will we still be able to build in a disconnected environment (brew)?
> 
> if we publish to npm registry, we will still add that dependency in yarn v2 dependencies of this downstream repository so it'll be there like any other javascript dependency

Okay, I've filed: https://github.com/eclipse/che/issues/19741


> > Well what takes the longest time is the che-theia-plugins.yaml as it's not a 1:1 subset of upstream, but that has to be done for both cases anyway. I've already started and should be done by this afternoon. Everything else is just copy/paste with a few small changes.
> 
> here I think it could be done automatically (improvement to made) to just have stuff automatically updated from the upstream yaml file

Yes, we'll need to do a general sync as there are a few things missing (for example upstream uses a mix of self-built and openvsx hosted vsixes). After that it should be easy :+1:

ericwill:
Things that will need to be handled in a separate PR:
* a script for changing the image tags from `:nightly` to `:VERSION` whenever the `VERSION` branch is cut
    * for example `:nightly` -> `:2.9` when the `crw-2.9-rhel8` branch is created
* a Dockerfile and copying mechanism for the vsix files, since they can't be downloaded in brew
* we'll probably need to convert this into a zero-installs build

nickboldt:
If you're happy to merge this as is, I am happy to see what happens... :)

As to the follow on work:

>  a script for changing the image tags from :nightly to :VERSION whenever the VERSION branch is cut, for example :nightly -> :2.9 when the crw-2.9-rhel8 branch is created

See https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/product/updateVersionAndRegistryTags.sh#L76-L92

> a Dockerfile and copying mechanism for the vsix files, since they can't be downloaded in brew
we'll probably need to convert this into a zero-installs build

Can we reuse/update https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/che-plugin-registry/build/scripts/cache_artifacts.sh ?

PR #483, opened 2021-05-03
fix(plugin-registry-entrypoint): Fix image regex should support single quote, multi line and multiple occurrences

sunix:
### What does this PR do?
1. Fix problems with image regexp:
   - single quotes not supported
   - multiline not supported
   - multiple occurence
2. Adding unit tests script

Applying upstream: https://github.com/eclipse-che/che-plugin-registry/pull/937

### DIFF
--- a/dependencies/che-plugin-registry/build.sh
+++ b/dependencies/che-plugin-registry/build.sh
@@ -10,6 +10,8 @@
 
 set -e
 
+base_dir=$(cd "$(dirname "$0")"; pwd)
+
 REGISTRY="quay.io"
 ORGANIZATION="crw"
 TAG="nightly"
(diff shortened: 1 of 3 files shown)


nickboldt (commented):
should contents of build/dockerfiles/entrypoint.sh be reordered so it matches the order in upstream PR? https://github.com/eclipse-che/che-plugin-registry/pull/937/files?w=1#diff-98eaea03c4a02f5f5f28cd0c4e1b1963732f47b3da641def607b71df56f751ae

nickboldt (approved):
+1, if Eric's cool with this.

ericwill (approved):

PR #487, opened 2021-05-06
Use zero-installs mode

ericwill:

### What does this PR do?
Use yarn zero installs so that deps don't need to be pulled in at build time.

Part of CRW-1648.


nickboldt:
Size of repo on disk when cloned:

```
git clone git@github.com:redhat-developer/codeready-workspaces.git
cd codeready-workspaces/; du -shc .

96M	total
```
vs.
```
co https://github.com/redhat-developer/codeready-workspaces/pull/487/files
du -shc .

169M	total
```

So, that's 70% larger. What's the speed boost we get here?

nickboldt (approved):
https://y.yarn.co/280c723e-ce81-4635-9e65-7102d00b3237_text.gif

PR #489, opened 2021-05-12
Prepare for Brew builds

ericwill:

### What does this PR do?
* Sync with latest upstream to include --offline build fixes
* Add npm/yarn installation in rhel.install script
* Fixed Dockerfile so that it can be used in both here and Brew
* Renamed images to proper registry.redhat.io ones

This is prerequisite PR to changes that will be coming in codeready-workspaces-images.

Part of CRW-1648

### DIFF
--- a/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
+++ b/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
@@ -10,19 +10,68 @@
 #   Red Hat, Inc. - initial API and implementation
 #
 
-# Build registry
+# Builder: check meta.yamls and create index.json
+# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8/python-38
+FROM registry.access.redhat.com/ubi8/python-38:1-54.1618436884 as builder
+USER 0
+
+ARG BOOTSTRAP=false
+ENV BOOTSTRAP=${BOOTSTRAP}
+ARG USE_DIGESTS=false
+ENV USE_DIGESTS=${USE_DIGESTS}
+
+# to get all the python deps pre-fetched so we can build in Brew:
+# 1. extract files in the container to your local filesystem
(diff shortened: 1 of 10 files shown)


dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -10,19 +10,68 @@
...
+RUN ./swap_images.sh output/v3
+RUN ./swap_plugins_memory.sh output/v3
+RUN if [[ ${USE_DIGESTS} == "true" ]]; then ./write_image_digests.sh output/v3; fi
+RUN ./list_referenced_images.sh output/v3 > /build/output/v3/external_images.txt
+RUN chmod -R g+rwX /build
+
+# Build registry, copying meta.yamls and index.json from builder
 # UPSTREAM: use RHEL7/RHSCL/httpd image so we're not required to authenticate with registry.redhat.io
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/rhscl/httpd-24-rhel7
-FROM registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-143.1619800879
```
nickboldt, line 16: this update to the container version will be reapplied when we sync later (assuming the upstream is ALSO updated to the latest base image)

dependencies/che-plugin-registry/che-editors.yaml
```diff
@@ -22,7 +22,7 @@ editors:
     components:
       - name: theia-ide
         container:
-          image: 'quay.io/crw/che-theia:nightly'
+          image: 'registry.redhat.io/codeready-workspaces/theia-rhel8:2.9'
```
nickboldt, line 25: I hope all these transforms were done via script, not manually :) and that the tag is a parameter

ericwill: does find/replace all + regex still count as manual? :laughing:

PR #495, opened 2021-05-17
Plugin registry script updates

ericwill:

### What does this PR do?
* Copy che-*.yaml files into Dockerfile
* Fix list_referenced_images.sh to use new che-*.yaml files
* Fix swap_images.sh and list_yamls.sh to use new che-*.yaml files

Part of CRW-1824

### DIFF
--- a/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
+++ b/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
@@ -46,14 +46,15 @@ COPY ./build/dockerfiles/rhel.install.sh /tmp
 RUN /tmp/rhel.install.sh && rm -f /tmp/rhel.install.sh
 
 COPY ./build/scripts/*.sh /build/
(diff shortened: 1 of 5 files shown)


ericwill:
The script invoked [here](https://github.com/redhat-developer/codeready-workspaces-operator/blob/crw-2-rhel-8/build/scripts/insert-related-images-to-csv.sh#L58-L61) will need to change from using the non-existent `v3` folder to using just `./`

ericwill:
> The script invoked [here](https://github.com/redhat-developer/codeready-workspaces-operator/blob/crw-2-rhel-8/build/scripts/insert-related-images-to-csv.sh#L58-L61) will need to change from using the non-existent `v3` folder to using just `./`

Opened a PR for that which can be merged after this one: https://github.com/redhat-developer/codeready-workspaces-operator/pull/49

nickboldt (approved):
looks great, no complaints.

PR #504, opened 2021-06-02
Use only latest plugin definitions for che-theia and che-machine-exec

ericwill:

### What does this PR do?
Deletes extra definitions and uses `latest` only

Part of CRW-1648


dependencies/che-plugin-registry/che-editors.yaml
```diff
@@ -2,7 +2,7 @@ version: 2.0.0
 editors:
   - schemaVersion: 2.1.0
     metadata:
-      name: eclipse/che-theia/next
+      name: eclipse/che-theia/latest
```
nickboldt, line 5: should line 11 be changed to latest too?

ericwill: you are correct

nickboldt (approved):
+1

benoitf:
Some PR checks are red

ericwill:
> Some PR checks are red

CI is broken, need to investigate why still

benoitf:
EoF newline ?

nickboldt:
> EoF newline ?

```
[EOF Rule] File "dependencies/LATEST_IMAGES" has lint errors
[EOF Rule] File "dependencies/LATEST_IMAGES_COMMITS" has lint errors
[EOF Rule] File "dependencies/LATEST_IMAGES_DIGESTS.json" has lint errors
```

3 generated files. unrelated to this PR since they were unchanged by this pr. Can I add them to an ignore list? Why did we enable them anyway? what value are these EOF checks providing?

benoitf:
If they're committed they should still have proper EOF ( And if generated it's also easy to generate proper files)

The value is to avoid the diff warning that you have by default when there is no EOF newline

PR #506, opened 2021-06-03
fix: add v3/plugins/.htaccess file back into Dockerfile

ericwill:

### What does this PR do?
Make sure the .htaccess file is copied in, it went missing in the recent plugin registry automation changes.

Part of CRW-1886

### DIFF
--- a/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
+++ b/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
@@ -84,6 +84,7 @@ WORKDIR /var/www/html
 
 RUN mkdir -m 777 /var/www/html/v3
 COPY README.md .htaccess /var/www/html/
+COPY /v3/plugins/.htaccess /var/www/html/v3/plugins/
 COPY --from=builder /build/output/v3 /var/www/html/v3
 COPY ./build/dockerfiles/rhel.entrypoint.sh ./build/dockerfiles/entrypoint.sh /usr/local/bin/
 RUN chmod g+rwX /usr/local/bin/entrypoint.sh /usr/local/bin/rhel.entrypoint.sh && \


sleshchenko (approved):

JPinkney (approved):

sleshchenko:
side-comment, does CRW repos need Conventional Commits? Probably yes

benoitf:
is that images folder is also there ?
https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/che-plugin-registry/v3/images/eclipse-che-logo.png

ericwill:
> is that images folder is also there ?
> https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/che-plugin-registry/v3/images/eclipse-che-logo.png

I think you are replying to the wrong PR

benoitf:
@ericwill I reply to the correct PR, just that I think we'll miss another file

```
$ docker run --rm -it --entrypoint=bash  quay.io/crw/pluginregistry-rhel8:2.9-20 -c "ls -la /var/www/html/v3/images/"
```

there is no eclipse-che-logo there

nickboldt (approved):

benoitf:
it's this upstream line
https://github.com/eclipse-che/che-plugin-registry/blob/master/build/dockerfiles/Dockerfile#L26

ericwill:
> @ericwill I reply to the correct PR, just that I think we'll miss another file
> 
> ```
> $ docker run --rm -it --entrypoint=bash  quay.io/crw/pluginregistry-rhel8:2.9-20 -c "ls -la /var/www/html/v3/images/"
> ```
> 
> there is no eclipse-che-logo there

Ah okay. There is another PR for icons open already, we'll handle it there

nickboldt:
> side-comment, does CRW repos need Conventional Commits? Probably yes

NEED? No. You can use semantic comments even if you're not forced to do so. 
If this PR was called "fix: ..." or "bug: ..." would you know significantly more about it than you do? 

I'd argue calling it "CRW-1886 ..." would be more valuable but I'm pretty good at forgetting to do that myself :)

But if someone wants to enable semantic commit pr checking here, have at it.

benoitf (approved):
we're missing another icon file but it'll handle in another PR

PR #509, opened 2021-06-04
chore(plugin-registry-generator): adding new option to be used in dev mode to use public images instead of unreleased crw images

sunix:
### What does this PR do?
Allow the option --use-dev-images to be used to replace image prefixed with registry.redhat.io/codeready-workspaces with quay.io/crw

Developers could use this command to build: 
```
BUILDER=docker SKIP_TEST=true SKIP_FORMAT=true SKIP_LINT=true  ./build.sh --skip-oci-image --use-dev-images
```
and safely try their plugin registry.

An example of a registry built with that command:
https://sunix-crw-plugin-registry-dev-h9894.surge.sh/v3/plugins/

Selection_383

### DIFF
--- a/dependencies/che-plugin-registry/tools/build/src/build.ts
+++ b/dependencies/che-plugin-registry/tools/build/src/build.ts
@@ -30,6 +30,7 @@ import { CheTheiaPluginsYamlGenerator } from './che-theia-plugin/che-theia-plugi
 import { CheTheiaPluginsYamlWriter } from './che-theia-plugin/che-theia-plugins-yaml-writer';
 import { Deferred } from './util/deferred';
 import { DigestImagesHelper } from './meta-yaml/digest-images-helper';
+import { DevImagesHelper } from './meta-yaml/dev-images-helper';
 import { ExternalImagesWriter } from './meta-yaml/external-images-writer';
 import { FeaturedAnalyzer } from './featured/featured-analyzer';
 import { FeaturedWriter } from './featured/featured-writer';
@@ -72,6 +73,10 @@ export class Build {
   @named('SKIP_DIGEST_GENERATION')
   private skipDigests: boolean;
 
+  @inject('boolean')
+  @named('USE_DEV_IMAGES')
+  private useDevImages: boolean;
+
   @inject(FeaturedAnalyzer)
   private featuredAnalyzer: FeaturedAnalyzer;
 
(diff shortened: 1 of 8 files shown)


dependencies/che-plugin-registry/tools/build/src/build.ts
```diff
@@ -320,11 +328,19 @@ export class Build {
     const computedYamls = [...cheTheiaPluginsMetaYaml, ...cheEditorsMetaYaml, ...chePluginsMetaYaml];
 
     let allMetaYamls = computedYamls;
+    if (this.useDevImages) {
```
benoitf, line 331: I would find a different way as the idea is to have like 100% code being upstream (and then even use the che-plugin-registry tool itself)

so probably we could have upstream a 'regexp/sed/replace' settings to apply on images

and in downstream we would only provide the parameter.

Else by merging this PR, codebase will differ

sunix: yep i was thinking backporting upstream.

nickboldt:
8 files changed and 100s of lines of code when a single line of sed would have achieved the same thing. +0 for this, especially this close to CRW 2.9 GA freeze. 

I'd be happier with this for 2.10

nickboldt:
Another approach would be to pass a flag to https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/che-plugin-registry/build/scripts/find_image_alternate_urls.sh so that instead of using internal brew images, we use external quay images.

Then you would be able to use the existing transformation code/logic already in the dockerfile to handle the case where we need to fetch digests for unreleased images.

sunix:
@nickboldt I don't think the last proposal is the right approach:
- It is not documented: we don't know the context, anyone coming from outside won't know how and when these script should be used
- It's fragile, no unit tests, if i pass a different parameter, i have no guarantees that these will work.
- So I read your flow many times but i found this not safier and equally complex compared to this PR
- Size is not that big if you ignore the tests folder.

nickboldt (changes requested):
Eric and Florent have both requested this change be done upstream first. You also have 4 failing tests. 

Therefore, -1.

nickboldt:
three weeks w/o an update. Closing - feel free to reopen if there's more work happening here in downstream (or link to your changes upstream? )

PR #515, opened 2021-06-08
fix(sed): update logic in write_image_digests.sh to handle single quotes

ericwill:

### What does this PR do?
Handle single quotes which are now present in all the image references.

CRW-1901


nickboldt:
Yay for yaml being inconsistent w/ its use of ', " and no quotes for attributes.

nickboldt:
merged in both 2 and 2.9 branches, thanks!

PR #525, opened 2021-06-15
fix(java): make java8 plugin executable

svor:

### What does this PR do?
Add a command to run executable endpoint for redhat/java8 plugin

https://issues.redhat.com/browse/CRW-1923

### DIFF
--- a/dependencies/che-plugin-registry/che-theia-plugins.yaml
+++ b/dependencies/che-plugin-registry/che-theia-plugins.yaml
@@ -327,6 +327,10 @@ plugins:
       memoryLimit: 1500Mi
       cpuLimit: 500m
       cpuRequest: 30m
+      args:
+        - sh
+        - -c
+        - ${PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}
       volumeMounts:
         - name: m2
           path: /home/jboss/.m2


nickboldt:
is this change required for other plugins too? Why is only java8 impacted?

svor:
> is this change required for other plugins too? Why is only java8 impacted?

@nickboldt because there is a difference between java8 plugin in [2.9](https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/che-plugin-registry/che-theia-plugins.yaml#L337) and [2.8](https://github.com/redhat-developer/codeready-workspaces/blob/crw-2.8-rhel-8/dependencies/che-plugin-registry/v3/plugins/redhat/java8/0.63.0/meta.yaml), probably it was missed when we made plugin's refactoring

nickboldt (approved):
seems legit. If Eric has no objections, then feel free to merge this into 2.9 and 2 branches, and trigger a 2.9 build (2.x jobs are currently disabled to save QE PSI testing resources).

benoitf:
about that change, as we need to apply the changes on all che-theia-plugins sidecars (as we reuse same image in devfile and plug-ins) maybe we need an extra flag on plug-in registry generator to add that on all sidecars.
(I see no value adding that in every entries of che-theia-plugins.yaml and it's leading to maintenance errors)

svor:
> about that change, as we need to apply the changes on all che-theia-plugins sidecars (as we reuse same image in devfile and plug-ins) maybe we need an extra flag on plug-in registry generator to add that on all sidecars.
> (I see no value adding that in every entries of che-theia-plugins.yaml and it's leading to maintenance errors)

@benoitf @nickboldt @ericwill I've created an issue to move this functionality from che-theia-plugins.yaml https://issues.redhat.com/browse/CRW-1924

PR #526, opened 2021-06-17
Fix: ShellCheck warnings, local image detection

themr0c:
### What does this PR do?

* Fix ShellCheck warnings
* Fix local image detection
* Fail when the image doesn't exist after pulling


themr0c:
For the local image detection, I could test only with `podman`. I didn't test with `docker`.

nickboldt:
replaced by #527

PR #533, opened 2021-06-23
fix(digests) do not rename .yaml files to .yaml.removed

ericwill:

### What does this PR do?
A failure in digest generation leads to all meta.yaml files in the registry being named *.yaml.removed. Let's not do this.

Part of CRW-1941


nickboldt:
rejected. this will cause devfiles / plugins that are not supported on some arches to re-appear (eg., dot net and fuse on z/p)

PR #545, opened 2021-07-14
feat: eclipse/che#18831 - Add option for updateBaseImages.sh to enforce PR creation

mkuznyetsov:

### What does this PR do?
Add option to enforce creation of PR when updating base images, to prevent unwanted direct push to target main branch

https://github.com/eclipse/che/issues/18831

### DIFF
--- a/product/updateBaseImages.sh
+++ b/product/updateBaseImages.sh
@@ -48,6 +48,7 @@ PR_BRANCH="pr-update-base-images-$(date +%s)"
 OPENBROWSERFLAG="" # if a PR is generated, open it in a browser
 docommit=1 # by default DO commit the change
 dopush=1 # by default DO push the change
+dopronly=0 # by default DO NOT enforce pushing only through PR branches
 buildCommand="echo" # By default, no build will be triggered when a change occurs; use -c for a container-build (or -s for scratch).
 
 checkrecentupdates () {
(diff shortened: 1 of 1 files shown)


product/updateBaseImages.sh
```diff
@@ -261,24 +267,19 @@ for d in $(find ${WORKDIR}/ -maxdepth ${MAXDEPTH} -name ${DOCKERFILE} | sort -r)
...
-											hub pull-request -f -m "${lastCommitComment}
-
-${lastCommitComment}" -b "${BRANCHUSED}" -h "${PR_BRANCH}" "${OPENBROWSERFLAG}" || true 
-										else
-											echo "# Warning: hub is required to generate pull requests. See https://hub.github.com/ to install it."
-											echo -n "# To manually create a pull request, go here: "
-											git config --get remote.origin.url | sed -r -e "s#:#/#" -e "s#git@#https://#" -e "s#\.git#/tree/${PR_BRANCH}/#"
+								git commit -s -m "chore: Update from ${URL} to ${FROMPREFIX}:${LATESTTAG}" ${DOCKERFILE}
+								if [[ ${dopronly} -eq 1]]; then
+									if [[ ${dopush} -eq 1 ]]; then
```
nickboldt, line 272: remove this second if check; doesn't really mean anything here.

product/updateBaseImages.sh
```diff
@@ -109,6 +111,7 @@ while [[ "$#" -gt 0 ]]; do
     '-s') buildCommand="rhpkg container-build --scratch"; shift 0;;
     '-n'|'--nocommit'|'--no-commit') docommit=0; dopush=0; shift 0;;
     '-p'|'--nopush'|'--no-push') dopush=0; shift 0;;
+	'-pr'|'--pronly'|'--pr-only') dopronly=1; shift 0;;
```
nickboldt, line 114: don't need 3 flags for this. just use `--pr`

product/updateBaseImages.sh
mkuznyetsov: yeah, dunno, maybe no push + pr mode would be used for "testing" purposes?

product/updateBaseImages.sh
mkuznyetsov: ok, thx

product/updateBaseImages.sh
nickboldt: so you want a PR generated but without generating the PR? that would essentially make the whole if-block do nothing

PR #547, opened 2021-07-19
fix: Fix updateBaseImages.sh PR branch creation

mkuznyetsov:

### What does this PR do?
Fix the cases when PR branch wouldn't be properly created in updateBaseImages.sh

N/A

#### Release Notes

N/A

### DIFF
--- a/product/updateBaseImages.sh
+++ b/product/updateBaseImages.sh
@@ -173,15 +173,19 @@ testvercomp () {
 
 
 createPr() {
-	git checkout -b $1 || true
-	git merge $2
+	headBranch=$1
+	baseBranch=$2
+	git branch ${headBranch} || true
+	git checkout ${headBranch}
+	git merge ${baseBranch}
+	git push origin ${headBranch}
 	lastCommitComment="$(git log -1 --pretty=%B)"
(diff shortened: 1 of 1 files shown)


product/updateBaseImages.sh
```diff
@@ -173,15 +173,19 @@ testvercomp () {
...
 
 createPr() {
-	git checkout -b $1 || true
-	git merge $2
+	headBranch=$1
+	baseBranch=$2
+	git branch ${headBranch} || true
+	git checkout ${headBranch}
+	git merge ${baseBranch}
+	git push origin ${headBranch}
```
nickboldt, line 181: do we need to push the topic branch explicitly? doesn't `hub pull-request` do this for us automatically?

product/updateBaseImages.sh
```diff
@@ -173,15 +173,19 @@ testvercomp () {
 
 
 createPr() {
-	git checkout -b $1 || true
-	git merge $2
+	headBranch=$1
+	baseBranch=$2
+	git branch ${headBranch} || true
```
nickboldt, line 178: this does the same as `checkout -b`

```
git checkout -b|-B <new_branch> [<start point>]
           Specifying -b causes a new branch to be created as if git-branch(1) were called and then checked out. 
```

PR #548, opened 2021-07-21
fix:CRW-1951 - Use a VS Code Language support for Camel extension version compatible with 8 months old VS Code API 1.50.0

apupier:
also removes the skipDependencies of vscode-commons which is useless as
there is no dependency from this extension to tvscode-commons

nota: it will remain VS Code Camel K that is not working but it requires more work so will be done in another PR

### What does this PR do?

Use a VS Code Language support for Camel compatible with 8 months old VS Code API 1.50.0

[CRW-1951](https://issues.redhat.com/browse/CRW-1951)


apupier:
using this other PR based on a branch in redhat-developr repository to avoid some job failures https://github.com/redhat-developer/codeready-workspaces/pull/550

PR #549, opened 2021-07-21
fix: CRW-1951 - use compatible version of VS Code Camel K to 8 months old VS code API 1.50.0

apupier:
it doesn't have the vscode-commons dependencies at that time
https://github.com/camel-tooling/vscode-camelk/blob/55155ec38a834ae75bacca6ff35a5ffefd6496af/package.json#L318-L321

Attempt withotu modifying the Camel K runtime provided in the sidecar. To be tested.


nickboldt:
Outdated. Using 0.0.26 in 2.12: https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/che-plugin-registry/che-theia-plugins.yaml#L169

PR #550, opened 2021-07-21
fix:CRW-1951 - Use a VS Code Language support for Camel extension version compatible with 8 months old VS Code API 1.50.0

apupier:
also removes the skipDependencies of vscode-commons which is useless as
there is no dependency from this extension to tvscode-commons

nota: it will remain VS Code Camel K that is not working but it requires more work so will be done in another PR

### What does this PR do?

Use a VS Code Language support for Camel compatible with 8 months old VS Code API 1.50.0

[CRW-1951](https://issues.redhat.com/browse/CRW-1951)

### DIFF
--- a/dependencies/che-plugin-registry/che-theia-plugins.yaml
+++ b/dependencies/che-plugin-registry/che-theia-plugins.yaml
@@ -375,7 +375,7 @@ plugins:
     extension: https://download.jboss.org/jbosstools/vscode/3rdparty/vscode-php-debug/php-debug-1.13.0.vsix
   - repository:
       url: 'https://github.com/camel-tooling/camel-lsp-client-vscode'
-      revision: 0.0.31
+      revision: 0.0.28
     sidecar:
       image: "registry.redhat.io/codeready-workspaces/plugin-java11-rhel8:2.11"
       name: vscode-apache-camel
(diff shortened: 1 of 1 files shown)


svor (commented):
I don't see an error with VS Code API and  Apache Camel v0.0.31. Probably we don't need to downgrade it. 
Screenshot from 2021-07-21 12-57-06

apupier:
> I don't see an error with VS Code API and Apache Camel v0.0.31. Probably we don't need to downgrade it

that's a good news. i'm interested to know in which Theia/Che/CRW, the VS Code API check is required and in which it is not to avoid any surprises

benoitf:
Also the check can be overrided by adding VSCODE_API_VERSION to another version than 1.50.0

svor:
> Also the check can be overrided by adding `VSCODE_API_VERSION` env var to another version than 1.50.0

@benoitf this env should be set into the container with che-theia?

benoitf:
@svor it can be done in the sidecar container (it's where the extension host will run)

svor (approved):

svor:
@apupier should we merge it?

nickboldt:
Note: I've closed related PR https://github.com/redhat-developer/codeready-workspaces/pull/549 so I'm inclined to close this one too, as it's been stalled for some time. 

If the fix is "done in the sidecar container" -- does that mean adding a new env var to https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-plugin-kubernetes/Dockerfile#L17-L19 instead of this PR?

nickboldt:
closing

PR #551, opened 2021-07-21
fix: remove skipDependencies on vscode-commons as it is a required dependency

apupier:
as I do no tknow why the skipDependencies awas introduced exatly and there is no other extensions that have it and really need it. try to remove the skip. See https://issues.redhat.com/browse/CRW-1951?focusedCommentId=16450102&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16450102

### DIFF
--- a/dependencies/che-plugin-registry/che-theia-plugins.yaml
+++ b/dependencies/che-plugin-registry/che-theia-plugins.yaml
@@ -167,9 +167,6 @@ plugins:
       cpuLimit: 500m
       cpuRequest: 30m
     extension: https://download.jboss.org/jbosstools/vscode/stable/vscode-camelk/vscode-camelk-0.0.24-198.vsix
-    metaYaml:
-      skipDependencies:
-        - redhat/vscode-commons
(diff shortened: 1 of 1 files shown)


apupier:
which seems not possible...

```Error: Unable to find the dependency id redhat/vscode-commons required by plug-in redhat/vscode-camelk```

unless we had the vscode-commons in the registry.

svor:
> which seems not possible...
> 
> `Error: Unable to find the dependency id redhat/vscode-commons required by plug-in redhat/vscode-camelk`
> 
> unless we had the vscode-commons in the registry.

looks like `redhat/vscode-camelk` v0.0.24 only one extension in CRW plugin-registry which depends on vscode-commons. And it should be skip by `skipDependencies` to build the registry

apupier:
closing fo rnow. i think upgrading to a newer version of VS Code Camel K or downgraded to a version compatible with VS Code API 1.50.0 are two better approaches than this one.

PR #557, opened 2021-07-30
fix: CRW-2118: Add httpd and ssl conf security updates 

nickboldt:
### What does this PR do?
Updates httpd to deny access to security sensitive files such as .htaccess
Updates ssl to increase encryption strength

https://issues.redhat.com/browse/CRW-2118

#### Release Notes

N/A

### DIFF
--- a/dependencies/che-devfile-registry/build/dockerfiles/Dockerfile
+++ b/dependencies/che-devfile-registry/build/dockerfiles/Dockerfile
@@ -8,6 +8,7 @@
 #
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
+#   IBM Corporation - implementation
 #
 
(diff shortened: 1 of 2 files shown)


nickboldt:
Upstream fixes in 

https://github.com/eclipse-che/che-devfile-registry/pull/459
https://github.com/eclipse-che/che-plugin-registry/pull/1014

svor (approved):

benoitf:
I don't see what this is fixing so it should probably be reverted

nickboldt:
Will watch for outcome in eclipse-che/che-devfile-registry#459 and eclipse-che/che-plugin-registry#1014.

If we need to revert this, let me know - can do so. Or @ArvinB can submit a PR to amend this change.

PR #564, opened 2021-08-06
fix: consume generator tool from upstream as an npm package

vinokurig:
Signed-off-by: Igor Vinokur <ivinokur@redhat.com>

### What does this PR do?
Consume generator tool from upstream as an npm package

https://issues.redhat.com/browse/CRW-1928


svor (approved):

dependencies/che-plugin-registry/build.sh
```diff
@@ -83,10 +83,8 @@ parse_arguments "$@"
 
 echo "Update yarn dependencies..."
 yarn
-echo "Build tooling..."
-yarn --cwd "$(pwd)/tools/build" build
 echo "Generate artifacts..."
-eval node "${NODE_BUILD_OPTIONS}" tools/build/lib/entrypoint.js --output-folder:"$(pwd)/output" "${BUILD_FLAGS_ARRAY[@]}"
+npx @eclipse-che/plugin-registry-generator@7.35.0-dev-01ff6be --root-folder:"$(pwd)" --output-folder:"$(pwd)/output" "${BUILD_FLAGS_ARRAY[@]}"
```
ericwill, line 87: If 2.11 is based on Che 7.34 we shouldn't be using upstream bits from Che 7.35. If necessary, let's backport upstream changes to `7.34.x` branch and publish the npm library from that. We can then consume that 7.34 version here.

ericwill (commented):
This is for 2.11 right? Shouldn't this PR be against the `crw-2.11-rhel8` branch then?

dependencies/che-plugin-registry/build.sh
nickboldt: This PR is for crw-2-rhel-8 = 2.12, so as it stands I'm OK with it. We'll need a new Prod Sec exception to be noted about this change, however, as it's a new npm dependency that we build (but which is hosted by 3rd party).

nickboldt (approved):
This PR is for crw-2-rhel-8 = 2.12, so as it stands I'm OK with it... though the JIRA says fixversion = 2.11, so we'll need a second PR for the 2.11 branch.

I also don't love that there are only "-dev-" versions in https://www.npmjs.com/package/@eclipse-che/plugin-registry-generator for the 2.11 release, I'd prefer to see a stable release at npmjs. Can we do some 7.yy.z releases in addition to the nightly devs?

We'll need a new Prod Sec exception to be noted about this change, however, as it's a new npm dependency that we build (but which is hosted by 3rd party). https://docs.google.com/document/d/1nw_O1QeG5bg1jfpoSkmABNiCjLpxxc9bSQXXLo8jgJw/edit#heading=h.ttxuwnu2u1jf

How will we ensure we're using the latest correct version of the package in 2.x and 2.11 branches? 

Can we add this version dependency into https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/VERSION.json and add some build process that pulls the version from there to make sure we build with the correct version?

dependencies/che-plugin-registry/build.sh
vinokurig: Opened another PR for the `2.11` branch https://github.com/redhat-developer/codeready-workspaces/pull/566

nickboldt: thx, looks good over in the other PR, except for the question about which z version to use...

nickboldt (changes requested):
Can you collect the version of the generator from a centalized place, eg.,

```
 ECPRG_VER=$(curl -sSLo- https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/crw-2-rhel-8/dependencies/VERSION.json \
| yq -r '.Other["@eclipse-che/plugin-registry-generator"]["2.12"]'); echo $ECPRG_VER
```

vinokurig:
@nickboldt 
> Can you collect the version of the generator from a centalized place, eg.,
>
>```
>ECPRG_VER=$(curl -sSLo- https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/crw-2-rhel->8/dependencies/VERSION.json \
>| yq -r '.Other["@eclipse-che/plugin-registry-generator"]["2.12"]'); echo $ECPRG_VER
>```

Done.

svor:
> ECPRG_VER=$(curl -sSLo- https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/crw-2-rhel-8/dependencies/VERSION.json \
| yq -r '.Other["@eclipse-che/plugin-registry-generator"]["2.12"]'); echo $ECPRG_VER

@nickboldt I don't think it is good to have hardcoded version in the script. But if we do that, shouldn't we use `2.x` for `crw-2-rhel-8` branch?

And it would be better to move that value into the varible in the script.

nickboldt:
> hardcoded version in the script. 

Agreed, better to pull the version from https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/VERSION.json#L2 

That way it'll automatically be updated to 2.13 when we move into the 2.12 branch and update 2.x to a newer future version

nickboldt:
ready to merge in both 2.11 and 2.x branches (change is now the same in both places and can simply compute its version and associated generator version, as long as the VERSION.json file in each branch is kept up to date.

nickboldt (approved):

nickboldt:
related: https://github.com/eclipse-che/che-plugin-registry/pull/1027

svor (approved):

benoitf (approved):

nickboldt:
Failed -- need to copy VERSION.json from upstream to midstream and move it into the root. 

https://github.com/redhat-developer/codeready-workspaces-images/commit/947db4159ac097ca45788d88b06493abb2edb3ee

Respinning: 
* https://main-jenkins-csb-crwqe.apps.ocp4.prod.psi.redhat.com/job/CRW_CI/job/crw-pluginregistry_2.11/11/
* https://main-jenkins-csb-crwqe.apps.ocp4.prod.psi.redhat.com/job/CRW_CI/job/crw-pluginregistry_2.x/204/

nickboldt:
Had to do a few more things for this to work, including moving to 0.0.2 version of che-openshift-auth plugin (as the 0.0.1 no longer exists). https://github.com/eclipse-che/che-plugin-registry/pull/1028

PR #566, opened 2021-08-10
fix: consume generator tool from upstream as an npm package

vinokurig:
Signed-off-by: Igor Vinokur <ivinokur@redhat.com>

### What does this PR do?
Consume generator tool from upstream as an npm package

https://issues.redhat.com/browse/CRW-1928


dependencies/che-plugin-registry/build.sh
```diff
@@ -83,10 +83,8 @@ parse_arguments "$@"
 
 echo "Update yarn dependencies..."
 yarn
-echo "Build tooling..."
-yarn --cwd "$(pwd)/tools/build" build
 echo "Generate artifacts..."
-eval node "${NODE_BUILD_OPTIONS}" tools/build/lib/entrypoint.js --output-folder:"$(pwd)/output" "${BUILD_FLAGS_ARRAY[@]}"
+npx @eclipse-che/plugin-registry-generator@7.34.0-dev-170a33f --root-folder:"$(pwd)" --output-folder:"$(pwd)/output" "${BUILD_FLAGS_ARRAY[@]}"
```
nickboldt, line 87: will this need to be updated to 7.34.2 when we release an update tomorrow?

nickboldt: Can you collect this value from a centalized place, eg.,

```
 ECPRG_VER=$(curl -sSLo- https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/crw-2-rhel-8/dependencies/VERSION.json | yq -r '.Other["@eclipse-che/plugin-registry-generator"]["2.11"]'); echo $ECPRG_VER
```

nickboldt (changes requested):
See comment.

dependencies/che-plugin-registry/build.sh
vinokurig: done

benoitf: https://www.npmjs.com/package/@eclipse-che/plugin-registry-generator/v/7.34.2

dependencies/che-plugin-registry/build.sh
```diff
@@ -83,10 +83,10 @@ parse_arguments "$@"
 
 echo "Update yarn dependencies..."
 yarn
-echo "Build tooling..."
-yarn --cwd "$(pwd)/tools/build" build
 echo "Generate artifacts..."
-eval node "${NODE_BUILD_OPTIONS}" tools/build/lib/entrypoint.js --output-folder:"$(pwd)/output" "${BUILD_FLAGS_ARRAY[@]}"
+ECPRG_VER=$(curl -sSLo- https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/crw-2.11-rhel-8/dependencies/VERSION.json | \
```
benoitf, line 87: BTW why use curl while the file is in the same repository ? so we don't have to care about the branch, just lookup at the file directly

dependencies/che-plugin-registry/build.sh
```diff
@@ -83,10 +83,10 @@ parse_arguments "$@"
...
+jq -r '.Other["@eclipse-che/plugin-registry-generator"]["2.11"]');
```
benoitf, line 88: 2.11 should not be harcoded as well `../dependencies/VERSION` file provides that data

dependencies/che-plugin-registry/build.sh
benoitf: I don't understand the name `ECPRG_VER`
We're not limited to characters so we may use a more explicit variable name
`UPSTREAM_VERSION` would be fine as well

dependencies/che-plugin-registry/build.sh
nickboldt: please don't use VERSION file, it's better to use the more robust VERSION.json file... eventually we may decide to scrap the single-serving VERSION file in favour of the more content-rich json one.

dependencies/che-plugin-registry/build.sh
nickboldt: eclipse-che/plugin-registry-generator => ECPRG 

But sure, call it whatever you want. That was just a first draft / suggestion to demo how you could read the content of the json to get the value you're hardcoding. 

(Since PRs are editable by contributors, you could just apply a change instead of offering discussion. Might be more efficient?)

dependencies/che-plugin-registry/build.sh
nickboldt: updated https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/VERSION.json#L462 to use 7.34.2

nickboldt:
ready to merge in both 2.11 and 2.x branches (change is now the same in both places and can simply compute its version and associated generator version, as long as the VERSION.json file in each branch is kept up to date.

nickboldt (approved):

svor (approved):

svor:
@benoitf any concerns regarding last changes?

benoitf (approved):

PR #574, opened 2021-08-23
feat: CRW-2155 Adding check to see if CRW_VERSION is in the json and …

SDawley:
…adds it based on the contents of 2.x if it is not.

### DIFF
--- a/product/updateVersionAndRegistryTags.sh
+++ b/product/updateVersionAndRegistryTags.sh
@@ -36,7 +36,6 @@ while [[ "$#" -gt 0 ]]; do
     '-b') BRANCH="$2"; shift 1;;
     '-v') CSV_VERSION="$2"; shift 1;; # 2.y.0
     '-t') CRW_VERSION="$2"; shift 1;; # 2.y
-    '-a') NEW_CRW_VERSION="$2"; shift 1;; 
     '-n'|'--no-commit') docommit=0; dopush=0; shift 0;;
     '-p'|'--no-push') dopush=0; shift 0;;
     '-prb') PR_BRANCH="$2"; shift 1;;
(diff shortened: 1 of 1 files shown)


product/updateVersionAndRegistryTags.sh
```diff
@@ -85,14 +84,45 @@ updateVersion() {
...
-    CRW_Y_VALUE="${CRW_VERSION#*.}"
-    UPPER_CHE=$(( (${CRW_Y_VALUE} + 6) * 2 ))
-    LOWER_CHE=$(( ((${CRW_Y_VALUE} + 6) * 2) - 1 ))
-
-    replaceField "${WORKDIR}/dependencies/VERSION.json" "(.Jobs[][\"${CRW_VERSION}\"]|select(.[]==\"main\"))" "[\"7.${UPPER_CHE}.x\",\"7.${LOWER_CHE}.x\"]"
-    replaceField "${WORKDIR}/dependencies/VERSION.json" "(.Jobs[][\"${CRW_VERSION}\"]|select(.[]==\"crw-2-rhel-8\"))" "[\"${BRANCH}\",\"${BRANCH}\"]"
+    
+    # CRW-2155, if version is in the json update it for che and crw branches
+    # otherwise inject new version.
+    check=$(cat ${WORKDIR}/dependencies/VERSION.json | grep "\"${CRW_VERSION}\"")
```
nickboldt, line 90: probably safe... but what if we happen to have some line with "2.13" in it because we're using traefik 2.13, and the check passes because of a false positive hit? Could we do this check with yq instead of grep? Should we?

SDawley: True, I went with grep after spending a lot of time trying to get jq to give me actionable results, but I'm willing to try again now that I know the keys command better.

Speaking of yq I should update those commands as well.

nickboldt:
+1 looks great

PR #578, opened 2021-08-30
chore: Dogfooding devfile for CRW devfile registry development

sunix:
This is a devfile based on the upstream one to build/run/publish(to surge.sh) the crw devfile registry.

### DIFF
--- a/dependencies/che-devfile-registry/devfile.yaml
+++ b/dependencies/che-devfile-registry/devfile.yaml
@@ -1,30 +1,69 @@
 apiVersion: 1.0.0
 metadata:
-  name: crw-devfileregistry
+  generateName: crw-devfile-registry-
 projects:
   - name: codeready-workspaces
     source:
       type: git
-      location: 'https://github.com/redhat-developer/codeready-workspaces.git'
+      location: 'https://github.com/redhat-developer/codeready-workspaces'
+  - name: che-devfile-registry
+    source:
+      type: git
+      location: 'https://github.com/eclipse/che-devfile-registry'
 components:
-  - alias: vscode-yaml
-    type: chePlugin
-    id: redhat/vscode-yaml/latest
-  - alias: registry-dev
-    type: dockerimage
-    image: 'quay.io/crw/devfileregistry-dev:latest'
+  - alias: che-devfile-registry-prod
+    image: 'quay.io/eclipse/che-devfile-registry:nightly'
     mountSources: true
-    memoryLimit: 1Gi
+    endpoints:
+      - name: http
+        port: 8080
+    command:
+      - sleep
+      - infinity
+    memoryLimit: 512Mi
+    type: dockerimage
+    volumes:
+      - name: images
+        containerPath: /usr/local/apache2/htdocs/images
+  - alias: dev
+    image: 'quay.io/eclipse/che-plugin-registry-dev:latest'
+    mountSources: true
+    memoryLimit: 1G
+    type: dockerimage
+  - id: redhat/vscode-xml/latest
+    type: chePlugin
+  - id: redhat/vscode-yaml/latest
+    type: chePlugin
+  - id: eamodio/gitlens/latest
+    type: chePlugin
+    registryUrl: 'https://che-plugin-registry-main.surge.sh/v3'
+  - id: donjayamanne/githistory/latest
+    type: chePlugin
+    registryUrl: 'https://che-plugin-registry-main.surge.sh/v3'
+  - id: timonwong/shellcheck/latest
+    type: chePlugin
+    registryUrl: 'https://che-plugin-registry-main.surge.sh/v3'
+  - id: ms-vscode/vscode-github-pullrequest/latest
+    type: chePlugin
 commands:
-  - name: '1. Validate devfiles'
+  - name: 1. build CRW devfile-registry
     actions:
-      - type: exec
-        component: registry-dev
-        command: './build/scripts/check_mandatory_fields.sh devfiles'
-        workdir: /projects/codeready-workspaces/dependencies/che-devfile-registry
-  - name: '2. list referenced images'
+      - workdir: '${CHE_PROJECTS_ROOT}/che-devfile-registry'
+        type: exec
+        command: REPO_DIR=/projects/codeready-workspaces/dependencies/che-devfile-registry ./build/dev/build-devfile-registry.sh
+        component: dev
+  - name: 2. start CRW devfile-registry
+    previewUrl:
+      port: 8080
+      path: /devfiles/
     actions:
-      - type: exec
-        component: registry-dev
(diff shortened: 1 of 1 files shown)


dependencies/che-devfile-registry/devfile.yaml
```diff
@@ -1,30 +1,69 @@
...
+    source:
+      type: git
+      location: 'https://github.com/eclipse/che-devfile-registry'
 components:
-  - alias: vscode-yaml
-    type: chePlugin
-    id: redhat/vscode-yaml/latest
-  - alias: registry-dev
+  - alias: che-devfile-registry-prod
+    image: 'quay.io/eclipse/che-devfile-registry:nightly'
```
nickboldt, line 15: why does the CRW devfile registry load the che-devfile-registry:nightly image? Also `:nightly` is dead; use `:next` for Che image tags.

dependencies/che-devfile-registry/devfile.yaml
```diff
@@ -1,30 +1,69 @@
...
+      - sleep
+      - infinity
+    memoryLimit: 512Mi
     type: dockerimage
-    image: 'quay.io/crw/devfileregistry-dev:latest'
+    volumes:
+      - name: images
+        containerPath: /usr/local/apache2/htdocs/images
+  - alias: dev
+    image: 'quay.io/eclipse/che-plugin-registry-dev:latest'
```
nickboldt, line 29: and here you're using the upstream `:latest` image tag. Again, why not the CRW image?

dependencies/che-devfile-registry/devfile.yaml
```diff
@@ -1,30 +1,69 @@
...
+        component: che-devfile-registry-prod
+  - name: 3. publish CRW devfile-registry to surge.sh
     actions:
-      - type: exec
-        component: registry-dev
-        command: './build/scripts/list_referenced_images.sh devfiles'
-        workdir: /projects/codeready-workspaces/dependencies/che-devfile-registry
+      - workdir: '${CHE_PROJECTS_ROOT}/che-devfile-registry/'
+        type: exec
+        command: REPO_DIR=/projects/codeready-workspaces/dependencies/che-devfile-registry ./build/dev/publish-devfile-registry-to-surge.sh
```
nickboldt, line 68: there is no such script in CRW repo. https://github.com/redhat-developer/codeready-workspaces/tree/crw-2-rhel-8/dependencies/che-devfile-registry/build contains no "dev" folder.

dependencies/che-devfile-registry/devfile.yaml
```diff
@@ -1,30 +1,69 @@
...
+  - id: redhat/vscode-xml/latest
+    type: chePlugin
+  - id: redhat/vscode-yaml/latest
+    type: chePlugin
+  - id: eamodio/gitlens/latest
+    type: chePlugin
+    registryUrl: 'https://che-plugin-registry-main.surge.sh/v3'
+  - id: donjayamanne/githistory/latest
+    type: chePlugin
+    registryUrl: 'https://che-plugin-registry-main.surge.sh/v3'
```
nickboldt, line 42: Using the upstream Che registries while dogfooding CRW means you'll be mixing in unsupported Che content into the CRW repo. Is that wise?

nickboldt (changes requested):
-1. I don't think this will work since the surge scripts don't exist in CRW repo. I'm also not sure why you're mixing Che stuff into the CRW build.

dependencies/che-devfile-registry/devfile.yaml
sunix: because it's a copy-paste :) I can rework it to use a CRW image yes

dependencies/che-devfile-registry/devfile.yaml
sunix: do we have `-dev` builder images in CRW ?

dependencies/che-devfile-registry/devfile.yaml
sunix: no but this is using the one from upstream, this devfile is cloning both project

dependencies/che-devfile-registry/devfile.yaml
sunix: These dogfooding devfiles would never be delivered to the customer. It is also supposed to be portable and we should be able to code CRW from a Che cluster.

dependencies/che-devfile-registry/devfile.yaml
nickboldt: Not as far as I know, other than theia-dev. You can see them in RHEC here: 
* https://catalog.redhat.com/software/containers/search?q=codeready-workspaces
or in quay:
* https://quay.io/organization/crw

What is the purpose of -dev containers? We build everything in Brew (or commandline w/ podman).

dependencies/che-devfile-registry/devfile.yaml
nickboldt: Since the CRW registry is a fork of the upstream Che registry (different devfiles, different/subset of plugins/sidecars) why are you trying to mix up and downstream into a single dev env?

dependencies/che-devfile-registry/devfile.yaml
nickboldt: > we should be able to code CRW from a Che cluster.
image

Since CRW is a fork, why would you ever start with Che to build CRW? Do you plan to include scripts to transform che registries into crw ones, like we do with operator and chectl/crwctl, stripping out functionality that doesn't exist downstream?

sunix: For the same reason some are using their local dev environment to test/contribute to this https://github.com/redhat-developer/codeready-workspaces.
It's like having a .vscode folder inside a che repo. It is about coding.

sunix: CRW is a cloud application like any others ... why can't I use Che to contribute to CRW ? in that case we should forbid people using VSCode, or IntelliJ ...

nickboldt: ok, that makese sense. As to `devfiles would never be delivered to the customer` I think we're OK there since they're not included in the devfile registry itself.

nickboldt:
Tempted to close this unmerged. @sunix will you be providing updates on this PR, or should we close it?

nickboldt:
Closing as no followup from @sunix :(

sunix:
I think this one would have been interesting if someone were using che or crw to contribute to the crw devfile registry. But if no one use it, let's skip it.

sunix:
also it would deserve a devfile2 migration

nickboldt:
If you want to contrib, happy to have it... as long as it's a complete devfile v2 solution :D

PR #592, opened 2021-09-17
Crw 2203 part2

SDawley:

### DIFF
--- a/product/updateVersionAndRegistryTags.sh
+++ b/product/updateVersionAndRegistryTags.sh
@@ -15,6 +15,7 @@ OPENBROWSERFLAG="" # if a PR is generated, open it in a browser
 docommit=1 # by default DO commit the change
 dopush=1 # by default DO push the change
 WORKDIR="$(pwd)"
+REMOVE_BRANCH=""
 
 usage () {
 	echo "Usage:   $0 -b [BRANCH] -v [CRW CSV VERSION] -t [CRW TAG VERSION] [-w WORKDIR]"
@@ -24,6 +25,7 @@ usage () {
 	--no-push, -p      do not push to BRANCH
 	-prb               set a PR_BRANCH; default: pr-update-version-and-registry-tags-(timestamp)
 	-o                 open browser if PR generated
+  -u [CRW VERSION]   remove data for [CRW VERSION] from job-config.json
 	--help, -h         help
 	"
 }
@@ -40,6 +42,7 @@ while [[ "$#" -gt 0 ]]; do
     '-p'|'--no-push') dopush=0; shift 0;;
     '-prb') PR_BRANCH="$2"; shift 1;;
     '-o') OPENBROWSERFLAG="-o"; shift 0;;
+    '-u') REMOVE_BRANCH="$2"; shift 1;;
     '--help'|'-h') usage; exit;;
     *) OTHER="${OTHER} $1"; shift 0;;
   esac
@@ -123,6 +126,30 @@ updateVersion() {
         fi
       done
     fi 
+
+    #remove unwanted version, if any0
+    if [[ $REMOVE_BRANCH ]]; then
+      TOP_KEYS=$(cat ${WORKDIR}/dependencies/job-config.json | jq 'keys')
+      TOP_KEYS=$(echo ${TOP_KEYS} | sed -e 's/\[//' -e 's/\]//' -e 's/\ //' -e 's/\,//g') #clean for array
+      TOP_KEYS=(${TOP_KEYS})
+
+      TOP_LENGTH=${#TOP_KEYS[@]}
+      for (( i=0; i<${TOP_LENGTH}; i++ ))
+      do
+        if [[ (${TOP_KEYS[i]} != "\"Version\"") && (${TOP_KEYS[i]} != "\"Copyright\"") && (${TOP_KEYS[i]} != "\"Purpose\"") ]]; then
+          # Get the sub-keys
+          KEYS=$(cat ${WORKDIR}/dependencies/job-config.json | jq '.'${TOP_KEYS[i]}' | keys')
+          KEYS=$(echo ${KEYS} | sed -e 's/\[//' -e 's/\]//' -e 's/\ //' -e 's/\,//g')
+          KEYS=(${KEYS})
+
+          KEYS_LENGTH=${#KEYS[@]}
+          for (( j=0; j<${KEYS_LENGTH}; j++ ))
+          do
+            replaceField "${WORKDIR}/dependencies/job-config.json" ".${TOP_KEYS[i]}[${KEYS[j]}]" "del(.\"${REMOVE_BRANCH}\")"
+          done
+        fi
+      done
+    fi
 }
 
 updateDevfileRegistry() {
--- a/product/updateVersionAndRegistryTags.sh
+++ b/product/updateVersionAndRegistryTags.sh
@@ -80,7 +80,23 @@ replaceField()
   fi
 }
 
-# update VERSION file to product version (x.y)
+updateBranchVersions()
+{
+  crwVersion="$1"
+  CRW_Y_VALUE="${crwVersion#*.}"
+  UPPER_CHE=$(( (${CRW_Y_VALUE} + 6) * 2 ))
+  LOWER_CHE=$(( ((${CRW_Y_VALUE} + 6) * 2) - 1 ))
+
+  UPPER_CHE_CHECK
+  if CHEEXISTS; then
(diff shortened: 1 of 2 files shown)


product/updateVersionAndRegistryTags.sh
```diff
@@ -40,6 +42,7 @@ while [[ "$#" -gt 0 ]]; do
     '-p'|'--no-push') dopush=0; shift 0;;
     '-prb') PR_BRANCH="$2"; shift 1;;
     '-o') OPENBROWSERFLAG="-o"; shift 0;;
+    '-u') REMOVE_BRANCH="$2"; shift 1;;
```
nickboldt, line 45: is this actually expecting a branch value (like 7.34.x or crw-2.11-rhel-8) or is it expecting a version (like 2.11 or 2.12) ? 

If the latter, maybe we want a variable like `REMOVE_CRW_VERSION` and a flag like `--remove` so it's more obvious what we're doing? (Aside: why "u" ? for "undo" ? )

product/updateVersionAndRegistryTags.sh
```diff
@@ -77,24 +80,27 @@ replaceField()
...
+# update job-config file to product version (x.y)
 updateVersion() {
     # deprecated, @since 2.11
     echo "${CRW_VERSION}" > "${WORKDIR}/dependencies/VERSION"
     # @since 2.11
     replaceField "${WORKDIR}/dependencies/job-config.json" '.Version' "${CRW_VERSION}"
     replaceField "${WORKDIR}/dependencies/job-config.json" '.Copyright' "[\"${COPYRIGHT}\"]"
+
+    CRW_Y_VALUE="${crwVersion#*.}"
+    UPPER_CHE=$(( (${CRW_Y_VALUE} + 6) * 2 ))
```
nickboldt, line 92: I might call these `UPPER_CHE_Y` and `LOWER_CHE_Y` since they're only the y value, not the 7.y value

product/updateVersionAndRegistryTags.sh
```diff
@@ -119,9 +125,51 @@ updateVersion() {
...
+        replaceField "${WORKDIR}/dependencies/job-config.json" "(.Jobs[][\"${crwVersion}\"]|select(.[]?==\"crw-2-rhel-8\"))" "[\"${BRANCH}\",\"${BRANCH}\"]"
+      fi
+
+      #make sure new builds are enabled
+      replaceField "${WORKDIR}/dependencies/job-config.json" "(.Jobs[][\"${CRW_VERSION}\"][\"disabled\"]|select(.==true))" 'false'
+      replaceField "${WORKDIR}/dependencies/job-config.json" "(.\"Management-Jobs\"[][\"${CRW_VERSION}\"][\"disabled\"]|select(.==true))" 'false'
+
+      #find and disable version-2
+      VERSION_KEYS=$(cat ${WORKDIR}/dependencies/job-config.json | jq '.Jobs'[\"dashboard\"]' | keys') #use dashboard as baseline since they should all be the same.
+      VERSION_KEYS=$(echo ${VERSION_KEYS} | sed -e 's/\[//' -e 's/\]//' -e 's/\ //' -e 's/\,//g')
```
nickboldt, line 157: if you're just trying to remove characters, could use tr here
```
$➔ echo "[foo, bar]" | tr -d "[] ," 
foobar
```

product/updateVersionAndRegistryTags.sh
```diff
@@ -119,9 +125,51 @@ updateVersion() {
...
+      VERSION_KEYS=($VERSION_KEYS)
+      OLD_VERSION=$(( ${#VERSION_KEYS[@]} - 4 ))
```
nickboldt, line 159: so... this is -4 because you're going from 7.38.x to 7.34.x ? 

what about disabling the jobs keyed to older crw-2.yy-rhel-8 branches?

product/updateVersionAndRegistryTags.sh
```diff
@@ -119,9 +125,51 @@ updateVersion() {
...
+         replaceField "${WORKDIR}/dependencies/job-config.json" "(.Jobs[][${VERSION_KEYS[$OLD_VERSION]}][\"disabled\"]|select(.==false))" 'true'
+         replaceField "${WORKDIR}/dependencies/job-config.json" "(.\"Management-Jobs\"[][${VERSION_KEYS[$OLD_VERSION]}][\"disabled\"]|select(.==false))" 'true'
+      fi
+
+      #update tags
+      replaceField "${WORKDIR}/dependencies/job-config.json" ".Other[\"FLOATING_QUAY_TAGS\"][\"${CRW_VERSION}\"]" "\"next\""
+
+      VERSION_LATEST=$(( ${#VERSION_KEYS[@]} - 3 )) 
+      replaceField "${WORKDIR}/dependencies/job-config.json" ".Other[\"FLOATING_QUAY_TAGS\"][${VERSION_KEYS[$VERSION_LATEST]}]" "\"latest\""
+      replaceField "${WORKDIR}/dependencies/job-config.json" ".Other[\"FLOATING_QUAY_TAGS\"][${VERSION_KEYS[$OLD_VERSION]}]" ${VERSION_KEYS[$OLD_VERSION]}
```
nickboldt, line 171: if we're adding/updating 2.13 and this is working on 2.11 (old_version), cool... but what version is in version_latest?

nickboldt: Also we have `OLD_VERSION` but `VERSION_LATEST` ... minor nitpick: could that be `LATEST_VERSION` for consistency with other `_VERSION` variables ?

product/updateVersionAndRegistryTags.sh
SDawley: Was expecting a version like 2.12 so `REMOVE_CRW_VERSION` is indeed the better option here. Originally went with -u like uninstall because `--remove` hadn't occurred to me...

product/updateVersionAndRegistryTags.sh
SDawley: So, what I'm actually doing is grabbing all the CRW_VERSION keys that are in json and making an array of those, then getting the minus forth element of the array and disabling everything of that version. So OLD_VERSION is the very poorly named array index of version - 2. This was my way of finding it without math in case the major version bumps so we go from say 2.15 to 3.0. 

From there though it should be pretty easy to make sure any earlier versions that are still around are also disabled.

product/updateVersionAndRegistryTags.sh
nickboldt: even simpler...
```
$➔ for f in $(jq -r '.Jobs'[\"dashboard\"]' | keys[]' job-config.json); do echo $f; done
2.11
2.12
2.13
2.x
````
or
`VERSION_KEYS=$(jq -r '.Jobs'[\"dashboard\"]' | keys[]' job-config.json)`

No need to sed or tr if the array braces and quotes are not there in the first place :)

File: product/updateVersionAndRegistryTags.sh (hunk 4 above)
Comment (nickboldt) [reply product/updateVersionAndRegistryTags.sh:159]: so you could do
```
VERSION_KEYS=($(jq -r '.Jobs'[\"dashboard\"]' | keys[]' job-config.json))
```

But what happens if the array only has 3 values: (2.11, 2.12, 2.x) or (2.11, 2.12, 2.13)

Can you do -4 if `${#VERSION_KEYS[@]}` = 3 ?

File: product/updateVersionAndRegistryTags.sh [hunk 6]
```diff
@@ -15,16 +15,18 @@ OPENBROWSERFLAG="" # if a PR is generated, open it in a browser
...
-	--no-commit, -n    do not commit to BRANCH
-	--no-push, -p      do not push to BRANCH
-	-prb               set a PR_BRANCH; default: pr-update-version-and-registry-tags-(timestamp)
-	-o                 open browser if PR generated
-	--help, -h         help
+	--no-commit, -n         do not commit to BRANCH
+	--no-push, -p           do not push to BRANCH
+	-prb                    set a PR_BRANCH; default: pr-update-version-and-registry-tags-(timestamp)
+	-o                      open browser if PR generated
+  --remove [CRW VERSION]  remove data for [CRW VERSION] from job-config.json
```
nickboldt, line 28: indent/alignment looks strange here. tabs v spaces?

nickboldt:
So the only remaining issue appears to be that when we add new versions, they're all tied to main, rather than 7.yy.x and 7.yy-1.x branches.

I did this:
```
./product/updateVersionAndRegistryTags.sh -n -b crw-2-rhel-8 -v 2.13.0 -t 2.13 -w `pwd`
./product/updateVersionAndRegistryTags.sh -n -b crw-2-rhel-8 -v 2.14.0 -t 2.14 -w `pwd`
./product/updateVersionAndRegistryTags.sh -n -b crw-2-rhel-8 -v 2.15.0 -t 2.15 -w `pwd`
```

and I don't see 2.14 or 2.15 mapped to 7.yy.x branches, only to main. 

Can we fix that?

PR #594, opened 2021-09-21
Adding content repo that are public facing

ArvinB:
These public facing content repos are for an image update. This allows downstream products to more easily adopt CRW. @nickboldt

### DIFF
--- a/dependencies/che-devfile-registry/build/dockerfiles/content_sets_ubi8.repo
+++ b/dependencies/che-devfile-registry/build/dockerfiles/content_sets_ubi8.repo
@@ -1,0 +1,20 @@
+[ubi-8-appstream]
+name=Red Hat Universal Base Image 8 (RPMs) - AppStream
+baseurl=https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/$basearch/appstream/os
+enabled=1
+gpgcheck=0
+skip_if_unavailable=True
+
+[ubi-8-baseos]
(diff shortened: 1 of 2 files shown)


ArvinB:
> Would it make sense to remove the disabled epel7 and redundant centos8 .repo files?
> 
> Also looking in the https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/dependencies/che-devfile-registry/build/dockerfiles/content_sets_rhel8.repo file I see there's no reference to Codeready builder repo... so you don't need it here.
> 
>     * https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-devfileregistry/content_sets.yml
> 
>     * https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-pluginregistry/content_sets.yml

* In regards to disabling `epel7`: I probably eventually would, but it's not going to hurt right now to leave it as-is.
* In regards to defining the Codeready builder: That is there for completeness. Even though its not used doesn't mean it may never be used. Also having the definition there is not going to have a negative effect. In fact, perhaps if an rpm is moved to the Codeready builder repo then having this already defined will be advantageous.

nickboldt:
test failed because from private fork

nickboldt:
Followup in https://github.com/redhat-developer/codeready-workspaces/pull/595

PR #601, opened 2021-10-05
Crw 2081

SDawley:
### What does this PR do?
Changes uploadAssetsToGHRelease.sh to use GitHub hub to check for/create/push to release. Adds installHub function to util.groovy. 

https://issues.redhat.com/browse/CRW-2081

### DIFF
--- a/product/uploadAssetsToGHRelease.sh
+++ b/product/uploadAssetsToGHRelease.sh
@@ -44,14 +44,13 @@ while [[ "$#" -gt 0 ]]; do
   shift 1
 done
 
-#check for hub and die if not there?
 export GITHUB_TOKEN=${GITHUB_TOKEN}
 
 # check if existing release exists
 #RELEASE_ID=$(curlWithToken -H "Accept: application/vnd.github.v3+json" $releases_URL | jq -r --arg PREFIX "${PREFIX}" --arg CSV_VERSION "${CSV_VERSION}" '.[] | select(.name=="Assets for the '$CSV_VERSION' '$PREFIX' release")|.url' || true); RELEASE_ID=${RELEASE_ID##*/}
-if [[ $(hub release | grep ${CSV_VERSION}) ]]; then
+if [[ !$(hub release | grep ${CSV_VERSION}) ]]; then
   #no existing release, create it
-  hub release create  -t "${MIDSTM_BRANCH}" -m "Assets for the ${CSV_VERSION} ${PREFIX} release" -m "Container build asset files for ${CSV_VERSION}" --prerelease "${CSV_VERSION}-${PREFIX}-assets"
+  hub release create -t "${MIDSTM_BRANCH}" -m "Assets for the ${CSV_VERSION} ${PREFIX} release" -m "Container build asset files for ${CSV_VERSION}" --prerelease "${CSV_VERSION}-${PREFIX}-assets"
 fi
 
 # upload artifacts for each platform 
--- a/product/uploadAssetsToGHRelease.sh
+++ b/product/uploadAssetsToGHRelease.sh
@@ -44,58 +44,28 @@ while [[ "$#" -gt 0 ]]; do
   shift 1
 done
 
-curlWithTokenPost()
-{
-  curl -sSL -XPOST -H "Authorization:token ${GITHUB_TOKEN}" "$1" "$2" "$3"
-}
-
-curlWithToken()
-{
-  curl -sSL -H "Authorization:token ${GITHUB_TOKEN}" "$1" "$2" "$3"
-}
-
-curlWithTokenBinary()
-{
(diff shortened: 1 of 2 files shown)


product/uploadAssetsToGHRelease.sh
```diff
@@ -44,58 +44,19 @@ while [[ "$#" -gt 0 ]]; do
...
-# shellcheck disable=2086
-RELEASE_ID=$(curlWithToken -H "Accept: application/vnd.github.v3+json" $releases_URL | jq -r --arg PREFIX "${PREFIX}" --arg CSV_VERSION "${CSV_VERSION}" '.[] | select(.name=="Assets for the '$CSV_VERSION' '$PREFIX' release")|.url' || true); RELEASE_ID=${RELEASE_ID##*/}
-if [[ -z $RELEASE_ID ]]; then 
-    curlWithTokenPost --data '{"tag_name": "'"${CSV_VERSION}-${PREFIX}-assets"'", "target_commitish": "'"${MIDSTM_BRANCH}"'", "name": "Assets for the '"${CSV_VERSION}"' '"${PREFIX}"' release", "body": "Container build asset files for '"${CSV_VERSION}"'", "draft": false, "prerelease": true}' $releases_URL > "/tmp/${CSV_VERSION}"
-    # Extract the id of the release from the creation response
-    RELEASE_ID="$(jq -r .id "/tmp/${CSV_VERSION}")"
+#RELEASE_ID=$(curlWithToken -H "Accept: application/vnd.github.v3+json" $releases_URL | jq -r --arg PREFIX "${PREFIX}" --arg CSV_VERSION "${CSV_VERSION}" '.[] | select(.name=="Assets for the '$CSV_VERSION' '$PREFIX' release")|.url' || true); RELEASE_ID=${RELEASE_ID##*/}
+if [[ !$(hub release | grep ${CSV_VERSION}) ]]; then
```
nickboldt, line 51: do you need a space between `!` and `$(...)` ?

product/uploadAssetsToGHRelease.sh
```diff
@@ -44,58 +44,19 @@ while [[ "$#" -gt 0 ]]; do
...
-        getId $RELEASE_ID $fileToPush
-        # TODO this doesn't work -- PATCH doesn't seem to support POSTing a new binary
-        if [[ $(curlWithTokenBinary @"${fileToPush}" -XPATCH "https://uploads.github.com/repos/redhat-developer/codeready-workspaces-images/releases/${RELEASE_ID}/assets/${ID}?name=${fileToPush}") ]]; then
-                getId $RELEASE_ID $fileToPush
-                echo "Updated asset $ID in https://api.github.com/repos/redhat-developer/codeready-workspaces-images/releases/$RELEASE_ID/assets/${ID}"
-        else
-            echo "ERROR: could not upload or update file $fileToPush to https://api.github.com/repos/redhat-developer/codeready-workspaces-images/releases/$RELEASE_ID/assets"
-        fi
-    fi
+    hub release edit -a ${fileToPush} "${CSV_VERSION}-${PREFIX}-assets"
```
nickboldt, line 60: does this require a `-m ""` flag?

```
      edit
           Edit the GitHub release for the specified TAG name. Accepts the same options as create command. Publish a draft with --draft=false.

           Without --message or --file, a text editor will open pre-populated with the current release title and body. To re-use existing title and body unchanged, pass -m "".
```

product/util.groovy
```diff
@@ -254,6 +254,16 @@ def installPodman(boolean usePulpRepos=false) {
   }
 }
 
+//compile and install github hub to push assets to release
+def installHub() {
```
nickboldt, line 258: Could pass in `String hubVersion = "2.14.2"` here so that if we wanted to, we could install a newer version easily... but maybe that's not needed atm.

product/uploadAssetsToGHRelease.sh
SDawley: I tested that logic in a shell script locally and it worked sans space

product/uploadAssetsToGHRelease.sh
SDawley: Yup, I'll add

product/util.groovy
```diff
@@ -254,6 +254,16 @@ def installPodman(boolean usePulpRepos=false) {
   }
 }
 
+//compile and install github hub to push assets to release
+def installHub(String hubVersion="2.14.2") {
```
nickboldt, line 258: I guess this now depends on installing golang

product/util.groovy
```diff
@@ -254,6 +254,37 @@ def installPodman(boolean usePulpRepos=false) {
   }
 }
 
+//compile and install github hub to push assets to release
+def installHub(String hubVersion="2.14.2", String goVersion="1.17.1", String minGoVersion, String arch) {
```
nickboldt, line 258: rather than having one version to install and one to check against, should just use the passed-in value as the one we check and install. therefore don't need minGoVersion, just goVersion, which will fall back to default 1.17.1 if not defined, but will otherwise be passed in from job-config.json -> groovy -> jenkinsfile..

PR #647, opened 2021-12-16
feat(devfile): Add a link to devfile v2 for microprofile-xp3 sample

svor:

### What does this PR do?
Adds a link to devfile version 2 for microprofile-xp3 sample. 
New devfile uses **mergeImage** sidecar-policy attribute, because we need `registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8` image to run WildFly server and `registry.redhat.io/codeready-workspaces/plugin-java11-rhel8` to run vscode-java plugin without problems.

The link to the devfile: https://github.com/crw-samples/microprofile-quickstart/blob/devfilev2/devfile.yaml

screenshot-codeready-openshift-workspaces apps cluster-93f6 93f6 sandbox925 opentlc com-2021 12 16-13_11_24

c2

screenshot-codeready-openshift-workspaces apps cluster-93f6 93f6 sandbox925 opentlc com-2021 12 16-13_35_24

https://issues.redhat.com/browse/CRW-2412

### DIFF
--- a/dependencies/che-devfile-registry/devfiles/00_java11-maven-microprofile-xp3/meta.yaml
+++ b/dependencies/che-devfile-registry/devfiles/00_java11-maven-microprofile-xp3/meta.yaml
@@ -3,3 +3,5 @@ displayName: Java with JBoss EAP XP 3.0 Microprofile
 description: Java stack with OpenJDK 11, Maven 3.6.2 and JBoss EAP XP 3.0
 tags: ["Java", "OpenJDK", "Maven", "EAP", "Microprofile", "EAP XP", "RHEL8"]
 icon: /images/type-jboss.svg
+links:
+  v2: https://github.com/crw-samples/microprofile-quickstart/tree/devfilev2


benoitf:
hello, could you tell what the problem is when using `registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8` to run the plug-ins ?

nickboldt:
> hello, could you tell what the problem is when using `registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8` to run the plug-ins ?

Where do you see a problem?

In https://github.com/redhat-developer/codeready-workspaces/blob/efd332ad72ff6cc790931d84b3df152b89bf8103/dependencies/che-devfile-registry/devfiles/00_java11-maven-microprofile-xp3/devfile.yaml#L26

we have `registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8:3.0-9` 

And in https://github.com/crw-samples/microprofile-quickstart/blob/devfilev2/devfile.yaml#L9

we have `registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8:3.0-9`

Of course this is **likely to break on power and z** because the sidecar is different on those arches, and while in the plugin reg we can [swap image references](https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-pluginregistry/build/scripts/swap_images.sh#L40-L41) when building for power and z... we can't do the saem for the same project's devfile.yaml as that's a single URL reference, not a built container.

benoitf:
@nickboldt not sure I follow your remark.
If an image is not supported on a specific arch it won't start in the workspace in any case.
With DevWorkspace, plug-ins can use the devfile image.

Also the link /devfilev2 is a link to the repository (so ok here it may not be the right image) but dashboard will use the prebuilt devWorkspaces (and not this link as it would not work for airgap, etc) so these templates can be modified for each arch to match the expected image.

nickboldt:
> @nickboldt not sure I follow your remark. If an image is not supported on a specific arch it won't start in the workspace in any case. With DevWorkspace, plug-ins can use the devfile image.

My point is that the power and z versions of the plugin registry refer to the openj9 versions of eap7 container. So they DO work in che-server / non-DW implementation, as the devfile is transformed to point to a valid sidecar. 

https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-devfileregistry/build/scripts/swap_images.sh#L33-L35

(sorry, I previously linked to the swap images script for plugin reg, not for devfile reg)

> Also the link /devfilev2 is a link to the repository (so ok here it may not be the right image) but dashboard will use the prebuilt devWorkspaces (and not this link as it would not work for airgap, etc) so these templates can be modified for each arch to match the expected image.

In this case the cached zip inside the devfile reg will contain same reference to the jdk container (not the j9 one) unless we have implemented the swap_images logic when collecting projects.

I"m not sure we have that in https://github.com/redhat-developer/codeready-workspaces-images/blob/crw-2-rhel-8/codeready-workspaces-devfileregistry/build/scripts/cache_projects.sh so it's likely that we need to add that or else we'll break all the java/eap devfiles for DW-enabled CRW 2.14 on Z and Power.

benoitf:
I don't understand why you're blocking this pull request

if `registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8` is not available on some arches it means that this example needs to be opt-out from these specific arches as it should be already the case.

benoitf:
and devWorkspace is a tech preview so better to be able to check if this sample is working rather than not being be able to test it unless it's there on all arches

nickboldt (commented):
Depending on TWO sidecars (registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8 and registry.redhat.io/codeready-workspaces/plugin-java11-rhel8) is going to hurt on Power and Z (see https://issues.redhat.com/browse/CRW-2578 ) but at least for x64, this PR seems reasonable.

benoitf:
for the first problem, about ? it's a 'long known issue' with openshift

It means that user has no entry in /etc/passwd 

Java is looking the home folder by looking at the /etc/passwd 'home' folder

benoitf:
about the second case, I think it's because the entrypoint is overrided

benoitf:
so EAP is not booted

benoitf:
so maybe adding as command the original entrypoint of the image might fix both the issues

svor:
> so maybe adding as command the original entrypoint of the image might fix both the issues

@benoitf I've tried to add a command to execute original entrypoint:
```
  - name: tools
    container:
      image: registry.redhat.io/jboss-eap-7/eap-xp3-openjdk11-openshift-rhel8:3.0-9
      command: ['/opt/eap/bin/openshift-launch.sh']
```
problem with running the application was solved
but plugins didn't start at all

benoitf:
@svor I think you need to use args so entrypoint of theia is used and then it delegates to the original one

benoitf (approved):
I'm not sure we should run anyway the tooling inside the EAP image so approving

benoitf:
post-Start event has been added to devWorkspace operator so it should solve the issue of being able to start the theia endpoint without overriding the entrypoint

In the meantime, we can stick as before with tooling running in a separate container

svor:
I'm going to merge this PR as it is

nickboldt:
cherrypicked to 2.14 branch too

PR #675, opened 2022-01-19
feat: Add default webhook trigger config method to util2.groovy

mkuznyetsov:


nickboldt (approved):
+1, nice use of closure passed into method!

PR #686, opened 2022-02-02
feat: add successful container build checker function

mkuznyetsov:

### What does this PR do?
Add utility method to check if rhpkg job has successfully built all containers (as total amount comes from container.yaml)

https://issues.redhat.com/browse/CRW-2195

### DIFF
--- a/product/util2.groovy
+++ b/product/util2.groovy
@@ -590,5 +590,18 @@ exit 0
   return statusCode > 1 ? false : true
 }
 
+boolean hasSuccessfullyBuiltAllPlatfroms(String containerYamlPath, String jobOutput) {
+  int containerBuildCount = sh(script: '''#!/bin/bash -xe
+    yq -r ".platforms.only | length" ''' + containerYamlPath, returnStdout: true).trim()
+    echo "Total container builds detected: "+containerBuildCount
+  int containerSuccessCount = jobOutput.count("build has finished successfully \\\\o/") - 1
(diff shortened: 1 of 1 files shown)


nickboldt:
per conversation we can probably move much of this into https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/product/rhpkg-container-build.sh

See also _test_ files like: the https://github.com/redhat-developer/codeready-workspaces/blob/crw-2-rhel-8/product/rhpkg-container-build.sh_test_consoleText_fail3_brew-build-failed.txt

product/util2.groovy
```diff
@@ -590,5 +590,18 @@ exit 0
   return statusCode > 1 ? false : true
 }
 
+boolean hasSuccessfullyBuiltAllPlatfroms(String containerYamlPath, String jobOutput) {
```
nickboldt, line 593: "hasSuccessfullyBuiltAllPlatfroms" ==> probably mean hasSuccessfullyBuiltAllPlatforms or hasSuccessfullyBuiltAllArches

nickboldt:
one problem with this check is that it won't update the `currentBuild.result`... if that's already been set to UNSTABLE or FAILURE, then this check won't change anything. :( 

So maybe we should do this check BEFORE filtering for errors/unknowns/failures, and if we get a pass with this check, skip the subsequent ones? 

Or do it after the error/unknown/failures, but if true, need to set `currentBuild.result` = SUCCESS.

PR #705, opened 2022-03-14
CRW-2591 - migrate Camel on springboot stack to java 11

apupier:
requires https://github.com/crw-samples/fuse-rest-http-booster/pull/4

### DIFF
--- a/dependencies/che-devfile-registry/devfiles/02_java11-maven-fuse/devfile.yaml
+++ b/dependencies/che-devfile-registry/devfiles/02_java11-maven-fuse/devfile.yaml
@@ -12,7 +12,7 @@ projects:
 components:
   -
     type: chePlugin
-    id: redhat/java8/latest
+    id: redhat/java11/latest
   -
     type: chePlugin
     id: redhat/vscode-apache-camel/latest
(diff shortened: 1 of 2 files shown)


nickboldt:
With this change, are s390x or ppc64le supported? If so, we can remove the zero-length files for arches that are now supported.

apupier:
> With this change, are s390x or ppc64le supported? If so, we can remove the zero-length files for arches that are now supported.

I think not directly.
The example migth need to be adjusted. Thi work has not been done in Red hat Intgeration https://issues.redhat.com/browse/ENTESB-17454?focusedCommentId=19578893&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-19578893 so I do not know.

it would require more investigation and testing than I'm unable to do.

As far as I know, the image mentioned here https://github.com/crw-samples/fuse-rest-http-booster/blob/edd52526643a4c51d5c5d7b9741d9394664704e6/pom.xml#L19 will need to be adjusted to use an IBm compatible version as I mentioned here https://issues.redhat.com/browse/ENTESB-17454?focusedCommentId=19529070&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-19529070
Also, as afar as I know, it was using openj9 and it is subject to be modified for next Fuse version because it wasn't compliant with legal so cannot be used yet.

PR #723, opened 2022-05-04
Add script for extracting DS/DWO/WTO from IIB image

amisevsk:
### What does this PR do?
Adds script `filterIIBForDevSpaces.sh` that can be used to pull DS/DWO/WTO and optionally CRW from an IIB image and build a new index image that contains only those operators. Note downloading the IIB and rendering it can take multiple minutes since it's so large.

The new index uses the exact same images as the IIB image, so it is suitable for testing Dev Spaces as you would from the IIB itself.

Usage:
```
./filterIIBForDevSpaces.sh \
  --iib registry-proxy.engineering.redhat.com/rh-osbs/iib:225370 \
  --tag test-index-image:dev
```

Results in
operatorhub-filtered
if this index is deployed as a CatalogSource

N/A

### DIFF
--- a/product/filterIIBForDevSpaces.sh
+++ b/product/filterIIBForDevSpaces.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+#
+# Copyright (c) 2018-2022 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


nickboldt:
I've created a new place for iibs to be pushed:
https://quay.io/repository/devspaces/iib?tab=tags

@amisevsk you should see an invite to join the devspaces quay org so you can push there too.

PR #727, opened 2022-05-10
chore: use internal url in plugin meta.yaml to download vsx

svor:

### What does this PR do?
Replaces `relative:extension` with plugin-registry internal URL at runtime in the plugin registry container

https://issues.redhat.com/browse/CRW-2973

### DIFF
--- a/dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
+++ b/dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
@@ -23,6 +23,7 @@ set -e
 REGISTRY=${CHE_SIDECAR_CONTAINERS_REGISTRY_URL}
 ORGANIZATION=${CHE_SIDECAR_CONTAINERS_REGISTRY_ORGANIZATION}
 TAG=${CHE_SIDECAR_CONTAINERS_REGISTRY_TAG}
+INTERNAL_URL=${CHE_PLUGIN_REGISTRY_INTERNAL_URL}
 
 DEFAULT_METAS_DIR="/var/www/html/v3"
 METAS_DIR="${METAS_DIR:-${DEFAULT_METAS_DIR}}"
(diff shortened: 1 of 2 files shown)


svor:
@nickboldt PR check image build is failed because it tries to pull `quay.io/devspaces/pluginregistry-rhel8:next-gh-action` image which doesn't exist.
Do you have an idea why we do `docker pull` before the build: https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/.github/workflows/plugin-registry-build.yaml#L68

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
```diff
@@ -225,6 +227,15 @@ function update_container_image_references() {
 
 }
 
+function update_extension_vsx_references() {
+    readarray -t metas < <(find "${METAS_DIR}" -name 'meta.yaml' -o -name 'devfile.yaml' -o -name 'che-theia-plugin.yaml')
```
benoitf, line 231: AFAIK it's theia endpoint that is replacing relative:extension by the internal registry URL 
https://github.com/eclipse-che/che-theia/blob/main/dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh#L71-L76

so I'm not sure what this code will do ?

svor: @benoitf https://github.com/eclipse-che/che-theia/blob/main/dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh#L71-L76 replaces `relative:extension` when the worksapce is starting , but we have a problem when trying to install plugins from the ide https://github.com/eclipse-che/che-theia/blob/main/extensions/eclipse-che-theia-remote-impl-k8s/src/node/k8s-plugin-service-impl.ts#L388-L397

PR #728, opened 2022-05-10
fix: remove redundant docker pull from GA

svor:

### What does this PR do?
Avoid cloning `quay.io/devspaces/pluginregistry-rhel8:next-gh-action` image before the image build to fix GitHub action, it seems this image is not needed for the build process.

### DIFF
--- a/.github/workflows/plugin-registry-pr-check-build.yaml
+++ b/.github/workflows/plugin-registry-pr-check-build.yaml
@@ -67,7 +67,6 @@ jobs:
         export SKIP_FORMAT=true
         export SKIP_LINT=true
         export SKIP_TEST=true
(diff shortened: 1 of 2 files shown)


.github/workflows/plugin-registry-pr-check-build.yaml
```diff
@@ -67,7 +67,6 @@ jobs:
         export SKIP_FORMAT=true
         export SKIP_LINT=true
         export SKIP_TEST=true
-        docker pull quay.io/devspaces/pluginregistry-rhel8:next-gh-action
```
nickboldt, line 70: definitely don't need this here! it's a different tag than what's used in `./build.sh`

PR #736, opened 2022-05-26
chore: update maven EAP devfile to use JDK11

svor:
### What does this PR do?
- update maven EAP devfile to use JDK11
- add maven EAP devfile for IBM Z/P


nickboldt:
cherrypicked to 3.0 branch as 66ef9534

PR #739, opened 2022-06-03
fix: remove meta.yaml if devfile.yaml.${arch} exists

svor:

### What does this PR do?
Marks meta.yaml as removed if devfile folder contains devfile.yaml.**arch** file. Where **arch** could be ppc64le or s390x.
When meta.yaml is marked as removed, the devfile should not be shown on the Create workspace view.

https://issues.redhat.com/browse/CRW-3034

### DIFF
--- a/dependencies/che-devfile-registry/build/scripts/swap_yamlfiles.sh
+++ b/dependencies/che-devfile-registry/build/scripts/swap_yamlfiles.sh
@@ -18,18 +18,11 @@ yamlfiles=$("$SCRIPT_DIR"/list_yaml.sh "$YAML_ROOT")
 
 # shellcheck disable=SC2086
 for yamlfile in $yamlfiles ; do
-  if [[ -e ${yamlfile}.${arch} ]] ; then
-      mv ${yamlfile} ${yamlfile}.orig
-      mv ${yamlfile}.${arch} ${yamlfile}
-      echo "[INFO] swapped to $arch version of ${yamlfile}.${arch}"
-  fi
-
-  # remove empty
-  if [[ ! -s ${yamlfile} ]] ; then
+  if [[ -e "$(dirname $yamlfile)/devfile.yaml.${arch}" ]] ; then
     mv ${yamlfile} ${yamlfile}.removed
     if [[ -e "$(dirname $yamlfile)/meta.yaml" ]] ; then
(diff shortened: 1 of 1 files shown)


nickboldt:
With this change we no longer have the ability to have a devfile that's different per arch, but I guess we already lost that when we moved to a single devfile.yaml in the remote repos, rather than the trio of devfiles in the che-devfile-registry folder. 

So, +1.

dependencies/che-devfile-registry/build/scripts/swap_yamlfiles.sh
```diff
@@ -18,18 +18,11 @@ yamlfiles=$("$SCRIPT_DIR"/list_yaml.sh "$YAML_ROOT")
...
 for yamlfile in $yamlfiles ; do
-  if [[ -e ${yamlfile}.${arch} ]] ; then
-      mv ${yamlfile} ${yamlfile}.orig
-      mv ${yamlfile}.${arch} ${yamlfile}
-      echo "[INFO] swapped to $arch version of ${yamlfile}.${arch}"
-  fi
-
-  # remove empty
-  if [[ ! -s ${yamlfile} ]] ; then
+  if [[ -e "$(dirname $yamlfile)/devfile.yaml.${arch}" ]] ; then
```
nickboldt, line 21: this is a "file exists" check, not a "file is empty" check.

Assuming we're ONLY going to have zero-length devfile.yaml.arch files to signal "not supported on this arch" and never have non-zero length replacement files (as we had in the past for cases where a change to the devfile would allow it to work on Z/Power w/ a different JDK).

But technically this is an API change from the old way of signalling an alternative/non-existant devfile v1 file. So... I wonder if we should have some sort of README somewhere so that anyone trying to build a custom devfile registry with ref to v2 devfiles won't hit this confusing situation?

svor: to avoid changing API we also have 2 possible solutions:
1. Add empty **devworkspace-che-theia-latest.yaml.ppc64le** and **devworkspace-che-theia-latest.yaml.s390x** into the folders with devfiles which are not supported in P/Z
2. Add empty **devfile.yaml** into 06_idea_rhel folder but we'll need to rework this API when we'll remove all devfile.yaml files https://issues.redhat.com/browse/CRW-2985

nickboldt (approved):

PR #747, opened 2022-06-20
chore: remove intellij sample from the registry

svor:

### What does this PR do?
Since we don't want to have samples which describe some special editor in the default devfile registry, IntelliJ IDEA Community Edition sample was removed from the devfile registry. It is still possible to use IntelliJ IDEA editor by following **Selecting an IDE** chapter from the documentation.

https://issues.redhat.com/browse/CRW-3014


nickboldt:
Note that the doc section in question is https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.0/html-single/user_guide/index#selecting-an-ide

l0rd (approved):

PR #748, opened 2022-06-23
WIP: CRW-3071, fixing inconsistencies in build script arguments

SDawley:
also variable names.

### What does this PR do?
Updates script arguments and variable names across repos, so we don't have -v for version and -t for tag doing the same thing. There will be corresponding changes in jeknins, images, chectl and theia.

https://issues.redhat.com/browse/CRW-3071

### DIFF
--- a/product/getLatestIIBs.sh
+++ b/product/getLatestIIBs.sh
@@ -12,10 +12,10 @@
 usage () {
 	echo "
 Usage: 
-  $0 -t PROD_VER [OPTIONS]
+  $0 -v PROD_VERSION [OPTIONS]
 
 Options:
-  -o OCP_VER          To limit results to a single OCP version, use this flag
+  -o OCP_VERSION          To limit results to a single OCP version, use this flag
   -p PROD_NAME        Defaults to 'Dev Spaces'; label on output when multiple OCP versions specified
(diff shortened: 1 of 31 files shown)


product/getLatestIIBs.sh
```diff
@@ -109,26 +109,25 @@ dsDefaults
...
-    '-c') CSVs="$2"; shift 1;;
-    '-i') IMAGE_PREFIX="$2"; shift 1;;
-    '-v') VERBOSE=1; QUIET="none"; shift 0;;
-    '-q'|'-qi') VERBOSE=0; QUIET="index"; shift 0;;
-    '-qb') VERBOSE=0; QUIET="bundle"; shift 0;;
-    '--crw') crwDefaults;;
-    '--ds')   dsDefaults;;
-    '--dwo') dwoDefaults;;
-    '--wto') wtoDefaults;;
+    '-v') PROD_VERSION="$2"; shift 2;;
```
nickboldt, line 112: the `-t` was intentional as it maps to the product version / *t*ag; maybe we want a different symbol for the product version/tag compared to the csv version? We could start using `--ver` and `--csv` so that `-v` and `-t` are only used for `verbose` and `target` ?

PR #773, opened 2022-07-26
chore: generate DevWorksapces with Code editor

svor:

### What does this PR do?
Generates devworkspaces (devworkspace-che-theia-latest.yaml) with Code editor for each sample from the devfile registry. 

As a result each sample folder should contain 3 files:
```
devworkspace-che-code-insiders.yaml  devworkspace-che-theia-latest.yaml  meta.yaml
```
Each sample in **index.json** looks like:
```json
{
    "displayName": "Go",
    "description": "Stack with Go",
    "tags": [
      "Tech-Preview",
      "Go",
      "Golang",
      "UBI8"
    ],
    "icon": "/images/type-go.svg",
    "links": {
      "v2": "https://github.com/devspaces-samples/golang-health-check/tree/devspaces-3-rhel-8",
      "devWorkspaces": {
        "eclipse/che-theia/latest": "/devfiles/05_go/devworkspace-che-theia-latest.yaml",
        "che-incubator/che-code/insiders": "/devfiles/05_go/devworkspace-che-code-insiders.yaml"
      }
    }
  }
``` 

To generate devworkspaces we use [che-code-devworkspace-handler](https://www.npmjs.com/package/@eclipse-che/che-code-devworkspace-handler) library. The source code of this lib is located [here](https://github.com/che-incubator/che-code/tree/main/tools/devworkspace-handler) 

https://issues.redhat.com/browse/CRW-3191

### DIFF
--- a/dependencies/che-devfile-registry/build.sh
+++ b/dependencies/che-devfile-registry/build.sh
@@ -80,9 +80,13 @@ function clear_generated_data() {
     rm -rf ./resources
     for dir in ./devfiles/*/
     do
-      FILE="${dir}/devworkspace-che-theia-latest.yaml"
-      if [[ -f "$FILE" ]]; then
-        rm "${FILE}"
+      CHE_THEIA_DW="${dir}/devworkspace-che-theia-latest.yaml"
+      if [[ -f "$CHE_THEIA_DW" ]]; then
+        rm "${CHE_THEIA_DW}"
+      fi
+      CHE_CODE_DW="${dir}/devworkspace-che-code-insiders.yaml"
+      if [[ -f "$CHE_CODE_DW" ]]; then
+        rm "${CHE_CODE_DW}"
       fi
     done
 }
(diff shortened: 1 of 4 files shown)


nickboldt:
So does this mean that a user can...
* select which IDE to use to open a given sample (new UI in dashboard), or
* see 2x as many sample tiles, and has to figure out which ones are for theia, and which ones are for vscode? Are they using labels like `community` or `tech preview` to identify which samples will open with which IDE? 

Also should we do the same for idea, the 3rd supported IDE?

svor:
For now Dashboard uses **eclipse/che-theia/latest** key to find devworksapce in **index.json** in the devfile registry to start a workspace. But when we decide using VSCode as a default editor, Dashboard will select a devworkspace with VSCode (**che-incubator/che-code/insiders**) to start a workspace. So, UI should not be changed and the number of samples is the same in the Dashboard.

I don't think we should do that for Idea, since we don't want to use it as a default editor for our samples.

nickboldt:
> UI should not be changed and the number of samples is the same in the Dashboard

+1, then.

PR #797, opened 2022-09-12
Introduce local Openvsx in the plugin registry

svor:
### What does this PR do?
Add local openvsx

notes:

- it uses gradle wrapper to build the sources of openvsx server
- postgresql was installed from https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm
- final user is **postgres** to make it possible to start postgresql

https://issues.redhat.com/browse/CRW-3295


dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -87,7 +109,26 @@ RUN sed -i /etc/httpd/conf/httpd.conf \
...
+    PGDATA=/var/lib/pgsql/14/data/database
+
+RUN mkdir -m 777 /openvsx-server
+# Copy OpenVSX server files
+COPY --from=openvsx-builder /tmp/openvsx-lib/ /openvsx-server
+# Copy OpenVSX configuration files
+COPY ./build/dockerfiles/application.yaml /openvsx-server/config/
+COPY ./build/dockerfiles/openvsx.conf /etc/httpd/conf.d/
+COPY ./build/dockerfiles/openvsx-sync.json /
+COPY ./build/dockerfiles/start-services.sh /
```
nickboldt, line 130: should these .sh files be in the root folder? why not /usr/bin or /opt or something?

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
```diff
@@ -154,50 +180,6 @@ function extract_and_use_related_images_env_variables_with_image_digest_info() {
                 fi
             done
         done
-    else
```
nickboldt, line 157: can you explain why this is no longer needed?

dependencies/che-plugin-registry/build/dockerfiles/openvsx-sync.json
```diff
@@ -0,0 +1,6 @@
+[
+    "ms-python.python",
```
nickboldt, line 2: is this a list of the plugins to include in the offline vsx registry? Should we include stuff for yaml/xml and the dependency analytics plugin(s) ?

dependencies/che-plugin-registry/build/scripts/build_openvsx.sh
```diff
@@ -0,0 +1,22 @@
...
+set -e
+set -o pipefail
+
+cd /tmp
+# temporary location
+git clone https://github.com/che-incubator/che-openvsx.git
+cd che-openvsx/server
+git checkout OPENVSX-498
+
+./gradlew --no-daemon assemble
```
nickboldt, line 16: how is this installed? and how will this work in Brew? Do we need to prefetch gradle as a zip and unpack it somewhere?

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,6 +11,16 @@
 #   IBM Corporation - implementation
 #
 
+# open-vsx builder
+FROM registry.access.redhat.com/ubi8/openjdk-11:1.14-3.1661798377 as openvsx-builder
```
nickboldt, line 15: I'm not sure if we can do multiple prebuild FROM stages in Brew.

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,6 +11,16 @@
...
+USER 0
+
+RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y git
+
+# It uses gradlew that downloads gradle to build sources
```
nickboldt, line 20:
```
$➔ podman run -it --rm --entrypoint /bin/bash registry.access.redhat.com/ubi8/openjdk-11

[jboss@6ed597d54b7f ~]$ gradle
bash: gradle: command not found
[jboss@6ed597d54b7f ~]$ gradlew
bash: gradlew: command not found
[jboss@6ed597d54b7f ~]$ whereis gradle
bash: whereis: command not found
[jboss@6ed597d54b7f ~]$ whereis gradlew
bash: whereis: command not found
[jboss@6ed597d54b7f ~]$ 
```

I suspect you need to install Gradle if you want to run it. :D

dependencies/che-plugin-registry/build/scripts/build_openvsx.sh
nickboldt: ah, I see we get gradle from https://github.com/che-incubator/che-openvsx/blob/master/server/gradle/wrapper/gradle-wrapper.properties#L3 ==> `https://services.gradle.org/distributions/gradle-6.9.1-bin.zip`

nickboldt: this might have to be rewritten downstream to use https://osbs.readthedocs.io/en/latest/users.html#fetch-artifacts-koji-yaml in container.yaml to pull the zip and unpack it

Also might have to figure out how to resolve all these deps https://github.com/che-incubator/che-openvsx/blob/master/server/build.gradle#L6

OR... we build / bootstrap all this in Jenkins, and publish the various files we need to GH, where we can then use fetch-artifacts to pull the files.

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
```diff
@@ -87,7 +87,7 @@ function run_main() {
       rm -rf /var/lib/pgsql/14/data/old
 
       # start postgres and openvsx
-      /start-services.sh
+      ./start-services.sh
```
nickboldt, line 90: this looks wrong, should probably be `/usr/local/bin/start-services.sh`

svor: yes, fixed it locally, but didn't push

svor:
The current version of image (it could be tested with vsvydenko/pluginregistry-rhel8:openvx) provides support for the local openvsx, public openvsx and local theia plugin registries:
screenshot-eclipse-che apps ci-ln-d7hhiit-72292 origin-ci-int-gce dev rhcloud com-2022 09 14-14_33_57
screenshot-eclipse-che apps ci-ln-rs9l92t-72292 origin-ci-int-gce dev rhcloud com-2022 09 12-21_52_44
screenshot-eclipse-che apps ci-ln-rs9l92t-72292 origin-ci-int-gce dev rhcloud com-2022 09 12-21_47_30

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
svor: in the environment with internet connection, gradlew downloads Gradle

dependencies/che-plugin-registry/build/dockerfiles/openvsx-sync.json
svor: yes, it's just a short one to build image faster. We'll definitely extend it

dependencies/che-plugin-registry/build/scripts/build_openvsx.sh
svor: Reworked it a bit. Now openvsx-server is built by [openvsx-builder.Dockrefile](https://github.com/redhat-developer/devspaces/pull/797/files#diff-a2f33fa2849573ed9458621c8f34d05172673471d7c4d41230b2aa36e5b4580c) and it creates an asset of openvsx server with vsix files

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
svor: reverted it back, am not sure if we still need this workaround, but it's another story

dependencies/che-plugin-registry/build.sh
```diff
@@ -75,6 +84,105 @@ function parse_arguments() {
 
 parse_arguments "$@"
 
+detectBuilder() {
```
nickboldt, line 87: does this need to be so complicated? why not just check for podman and fall back to docker, else fail? 

eg., 
``` 
if [[ ! $(command -v podman) ]] && [[ ! $(command -v docker) ]]; then 
    echo "Please install podman or docker to continue." ; exit 1;
fi
```

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,95 +11,120 @@
...
-    echo "</FilesMatch>" >> /etc/httpd/conf/httpd.conf
-
-RUN sed -i /etc/httpd/conf.d/ssl.conf \
-    -e "s,SSLProtocol all -SSLv2,SSLProtocol all -SSLv3," \
-    -e "s,SSLCipherSuite HIGH:MEDIUM:!aNULL:!MD5,SSLCipherSuite HIGH:!aNULL:!MD5,"
-
-RUN sed -i /etc/httpd/conf/httpd.conf \
+RUN mkdir -p /mnt/rootfs
+# Install httpd and postgresql
+RUN yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm
```
nickboldt, line 29: do we need this repo? it introduces the possibility of installing a DIFFERENT set of RPMs than what's used in Brew. (and it's not allowed in Brew)

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,95 +11,120 @@
...
+RUN yum install --installroot /mnt/rootfs postgresql13-libs postgresql13 postgresql13-server \
```
nickboldt, line 30: maybe we can do the same as the postgresql-13 image? see line 44 of https://catalog.redhat.com/software/containers/rhel8/postgresql-13/5ffdbdef73a65398111b8362?container-tabs=dockerfile

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
```diff
@@ -159,8 +185,8 @@ function extract_and_use_related_images_env_variables_with_image_digest_info() {
         # Try to read RELATED_IMAGES from devspaces.csv.yaml (this will not work in disconnected environment).
         # DS_BRANCH env descries the branch where related csv.yaml is located; 
         # default value is devspaces-3-rhel-8 but should be overwritten when built from a stable branch like devspaces-3.1-rhel-8
-        curl -sSLo csv.yaml https://raw.githubusercontent.com/redhat-developer/devspaces-images/"${DS_BRANCH}"/devspaces-operator-bundle-generated/manifests/devspaces.csv.yaml
-        readarray -t images < <(grep "image:" csv.yaml | sed -r "s;.*image:[[:space:]]*'?\"?([._:a-zA-Z0-9-]*/?[._a-zA-Z0-9-]*/[._a-zA-Z0-9-]*(@sha256)?:?[._a-zA-Z0-9-]*)'?\"?[[:space:]]*;\1;")
+        curl -o /tmp/csv.yaml https://raw.githubusercontent.com/redhat-developer/devspaces-images/"${DS_BRANCH}"/devspaces-operator-bundle-generated/manifests/devspaces.csv.yaml
```
nickboldt, line 188: you're likely going to want to put the `-sSL` back so that output is quieter and symlinks/redirects are followed. Was it causing a problem?

dependencies/che-plugin-registry/build/dockerfiles/ovsx-installer.Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+ENV PATH="/tmp/opt/nodejs/bin:$PATH" \
+    npm_config_cache=/tmp/opt/cache
+
+RUN yum install curl -y --nodocs
+
+USER 1001
+
+RUN mkdir -p /tmp/opt/nodejs && mkdir -p /tmp/otp/cache &&\
+    curl -sL https://nodejs.org/download/release/v14.18.3/node-v14.18.3-linux-x64.tar.gz | tar xzf - -C /tmp/opt/nodejs --strip-components=1
```
nickboldt, line 20: do we really need node 14? everything else in Che is node 16 now. Would be nice to use the same version if possible.

nickboldt: You could also install node like this:
https://catalog.redhat.com/software/containers/ubi8/nodejs-16/615aee9fc739c0a4123a87e1?container-tabs=dockerfile
(line 56, enable the module and install via RPM)

dependencies/che-plugin-registry/build/scripts/build_openvsx.sh
ArvinB: I recommend a gradle wrapper 7.x. This has enhancements that were implemented requiring a better plugin architecture and enhanced security.

dependencies/che-plugin-registry/build/scripts/start_services.sh
```diff
@@ -0,0 +1,19 @@
...
+# wait that postgresql is ready
+printf "Waiting that postgresql is ready"
+timeout 0 bash -c "until /usr/pgsql-13/bin/pg_isready -h 127.0.0.1 -p 5432 -U postgres -q; do printf '.'; sleep 1; done"
+echo "Database is ready"
+
+# start openvsx
+pushd /openvsx-server || return
+./run-server.sh &
+printf "Waiting that openvsx is ready"
+timeout 0 bash -c "until curl --output /dev/null --head --silent --fail http://localhost:9000/user; do printf '.'; sleep 1; done"
```
nickboldt, line 18: do ports 5432 and 9000 need to be exposed somehow to openshift for this to work in devspaces?

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,95 +11,120 @@
...
-    echo "Installed Packages" && rpm -qa | sort -V && echo "End Of Installed Packages"
-
-RUN echo "<FilesMatch "\""^\\.ht"\"">" >> /etc/httpd/conf/httpd.conf && \
-    echo "Require all denied" >> /etc/httpd/conf/httpd.conf && \
-    echo "</FilesMatch>" >> /etc/httpd/conf/httpd.conf
-
-RUN sed -i /etc/httpd/conf.d/ssl.conf \
-    -e "s,SSLProtocol all -SSLv2,SSLProtocol all -SSLv3," \
-    -e "s,SSLCipherSuite HIGH:MEDIUM:!aNULL:!MD5,SSLCipherSuite HIGH:!aNULL:!MD5,"
-
```
ArvinB, line 80: These lines existed to increase security for Apache. If httpd is no longer being used, then please ignore this.

dependencies/che-plugin-registry/build/scripts/build_openvsx.sh
nickboldt: can you suggest what to change @ArvinB ? Which 7.x do you recommend?

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
nickboldt: https is still used. So I guess we need to re-add these config steps @svor  ?

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,95 +11,120 @@
...
-RUN ./swap_plugins_memory.sh ./output/v3
-RUN ./list_referenced_images.sh ./output/v3 --use-generated-content > /build/output/v3/external_images.txt && cat /build/output/v3/external_images.txt
-RUN chmod -R g+rwX /build
-
-# Build registry, copying meta.yamls and index.json from builder
-# https://registry.access.redhat.com/ubi8/httpd-24
-FROM registry.access.redhat.com/ubi8/httpd-24:1-199 AS registry
-USER 0
-
-# latest httpd container doesn't include ssl cert, so generate one
```
nickboldt, line 65: please re-enable this step after installing the `httpd` rpm on line 31

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,95 +11,120 @@
...
-RUN chmod +x /usr/share/container-scripts/httpd/pre-init/40-ssl-certs.sh && \
-    /usr/share/container-scripts/httpd/pre-init/40-ssl-certs.sh
-RUN \
```
nickboldt, line 68: please re-enable this step after installing the last rpm, as we need this in the console log to create rpm installation manifests for prod sec

dependencies/che-plugin-registry/build/scripts/build_openvsx.sh
ArvinB: @nickboldt I'm using 7.3 for this past year with solid results. I use Gradle for our multi-arch container image builds. However, the latest version is 7.5.1, which I'm likely going to upgrade to here in the next week. Once you move to any 7.x version, anything from 6.x that is no longer supported...the gradle wrapper will let you know during a build.

```
gradle wrapper --gradle-version 7.5.1
```

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -11,95 +11,120 @@
...
+RUN tar -xvf /build/resources.tgz -C /build/
+RUN rm -rf /build/output/v3/che-theia-plugins.yaml /build/output/v3/che-editors.yaml /build/output/v3/che-plugins.yaml
+RUN /build/swap_plugins_memory.sh /build/output/v3
+RUN /build/list_referenced_images.sh /build/output/v3 --use-generated-content > /build/output/v3/external_images.txt && cat /build/output/v3/external_images.txt
+RUN chmod -R g+rwX /build
+
+RUN cp -r /build/output/v3 /mnt/rootfs/var/www/html/
+COPY v3/plugins/ /mnt/rootfs/var/www/html/v3/plugins/
+COPY v3/images/*.png /mnt/rootfs/var/www/html/v3/images/
+COPY ./build/dockerfiles/rhel.entrypoint.sh ./build/dockerfiles/entrypoint.sh /mnt/rootfs/usr/local/bin/
```
nickboldt, line 96: you copy both entrypoint.sh files here, but then don't use the rhel variant on line 103?

dependencies/che-plugin-registry/build/scripts/start_services.sh
benoitf: no, db port is internal and 9000 openvsx port is private as well (exposed only through apache)

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
svor: since now we have to run a container with non root user (needed for postgresql), it's a permission problem to use `-sSL` in `curl` command

dependencies/che-plugin-registry/build/dockerfiles/ovsx-installer.Dockerfile
svor: updated to node 16

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
benoitf: no, this should not
https is not enabled

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
benoitf: it should not be re-enabled
there is no ssl configuration
it only listen on the http port / 80 port

dependencies/che-plugin-registry/build/dockerfiles/ovsx-installer.Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+# install the ovsx cli
+RUN npm install -g ovsx@0.5.0 && chmod -R g+rwX /tmp/opt/nodejs
+RUN tar -czvf nodejs.tar.gz /tmp/opt/nodejs
+
+FROM registry.access.redhat.com/ubi8/ubi-micro:8.5-744
+WORKDIR /ovsx
+COPY --from=builder /opt/app-root/src/nodejs.tar.gz nodejs.tar.gz
+RUN chmod g+rwX /ovsx/nodejs.tar.gz
```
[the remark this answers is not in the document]
svor: yes, because we need to use **ovsx** npm library from this tarball to prepare openvsx registry

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
benoitf: there is no need to have ssl cert as there is no https

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
benoitf: no ssl stuff so no need

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
svor: removed rhel.entrypoint.sh

dependencies/che-plugin-registry/build.sh
svor: I've just moved a pies of the working code into the function

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
svor: it seems we need another tarball for postgresql

dependencies/che-plugin-registry/build/scripts/build_openvsx.sh
svor: we use the same version that is used in the [upstream](https://github.com/eclipse/openvsx/blob/master/server/gradle/wrapper/gradle-wrapper.properties#L3)

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
svor: @nickboldt I've changed the code to install postgresql from the local rpm, WDYT?

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
svor: provided postgresql-13 as a tarball

dependencies/che-plugin-registry/build/dockerfiles/ovsx-installer.Dockerfile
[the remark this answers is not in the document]
nickboldt: but you're not unpacking it here in the installer.Dockerfile. Since the purpose of this image appears to be to provide a tarball that we then remove from the container ... 

why not just remove it from `/opt/app-root/src/nodejs.tar.gz` and skip the need for the ubi-micro container and lines 25-29 entirely?

Also if you built it in the `builder` stage like this:

`RUN tar -czvf nodejs.tar.gz /tmp/opt/nodejs`

then how can you access it to copy it from `/opt/app-root/src/nodejs.tar.gz` ? 

should the line be:

`COPY --from=builder /opt/app-root/src/nodejs.tar.gz nodejs.tar.gz` ?

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
nickboldt: those just mean "silent", "show errors", and "follow links"

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
nickboldt: looks good, I like the new `postgresql13.tar.gz`.

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
nickboldt: yep, postgresql13.tar.gz

dependencies/che-plugin-registry/build/dockerfiles/ovsx-installer.Dockerfile
nickboldt: yep, hardcoded to 16.17.1 but https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/dependencies/job-config.json#L778 uses 16.13.2.

Would be ideal to have everything depend on the same value, but this is OK for now. https://issues.redhat.com/browse/CRW-3401 for followup enhancement to avoid things getting misaligned.

nickboldt (approved):
+1 from me, though I find it strange we're using http (not https) in 2022, even for internally served files. :shrug: I defer to @benoitf on that one. 

on to https://github.com/redhat-developer/devspaces-images/pull/333 for review...

benoitf:
@nickboldt AFAIK no internal service is served over https (theia, vscode, devfile registry or plugin-registry)

dependencies/che-plugin-registry/build/dockerfiles/entrypoint.sh
svor: Fixed. Using -sSL but with **/tmp**/csv.yaml

PR #826, opened 2022-10-17
Filter operators from multiple IIBs when copying IIBs to quay.

amisevsk:
### What does this PR do?
Updates `copyIIBsToQuay.sh` to use the Dev Spaces IIB for the Dev Spaces and Web Terminal Operators, and the latest DevWorkspaces IIB for the DevWorkspace Operator.

To do this, I split `filterIIBForDevSpaces.sh` into two scripts, one to do the filtering and another to do the building/pushing.

* Want to use latest RC builds of DWO for testing Dev Spaces.
* https://issues.redhat.com/browse/CRW-3330


product/filterIIBForPackage.sh
```diff
@@ -0,0 +1,157 @@
...
+
+Usage: $0 [OPTIONS]
+
+Options:
+  -s, --iib <source_index>           : Source registry, org, index image and tag from which to pull operators. Required.
+  --channel-ds  <channel_name>       : Target channel to use when publishing Dev Spaces; if not set, use same channel in
+                                       source IIB image (eg., stable)
+  --channel-all <channel_name>       : Target channel to use when publishing all operators
+  --packages '<package1> <package2>' : Space separated list of packages to filter and include in target image. If 
+                                       unspecified, default is 'devspaces devworkspace-operator web-terminal'
```
nickboldt, line 32: might consider listing these in install sequence rather than alpha order?

product/buildCatalogFromFiles.sh
```diff
@@ -0,0 +1,127 @@
...
+pushd $WORKING_DIR > /dev/null
+trap 'popd > /dev/null' EXIT
+
+if [ -f ./olm-catalog.Dockerfile ]; then rm -f ./olm-catalog.Dockerfile; fi
+$PODMAN rmi --ignore --force $targetIndexImage >/dev/null 2>&1 || true
+
+# old way for olm 4.10
+# opm alpha generate dockerfile ./olm-catalog
+
+# new way for old 4.11 - see https://docs.openshift.com/container-platform/4.11/operators/admin/olm-managing-custom-catalogs.html#olm-creating-fb-catalog-image_olm-managing-custom-catalogs
```
nickboldt, line 92:
```suggest
# new way for olm 4.11 - see https://docs.openshift.com/container-platform/4.11/operators/admin/olm-managing-custom-catalogs.html#olm-creating-fb-catalog-image_olm-managing-custom-catalogs
```

product/filterIIBForPackage.sh
```diff
@@ -0,0 +1,157 @@
...
+}
+
+VERBOSE=0
+PACKAGES='devspaces devworkspace-operator web-terminal'
+WORKING_DIR='./'
+
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    '-s'|'--iib') sourceIndexImage="$2"; shift 1;;
+    '--channel-ds') targetChannelDS="$2"; shift 1;;
```
nickboldt, line 54: I thought you were going to remove this flag and variable so we only use `--channel` instead of `--channel-ds` and `--channel-all` ?

product/buildCatalogFromFiles.sh
```diff
@@ -0,0 +1,127 @@
...
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Utility script build a catalog image from bundle, channel, and package files rendered
+# by filterIIBForPackage.sh. The built catalog contains only operators in files and is
+# thus smaller in size.
+# OPM 4.11 is required to run buildCatalogFromFiles.sh
```
nickboldt, line 13: naming question... is"FromFiles" meaningful? We could just call this "buildCatalog.sh" ...

product/filterIIBForPackage.sh
```diff
@@ -0,0 +1,157 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Utility script to filter the Dev Spaces, Web Terminal, DevWorkspace (and optionally CodeReady Workspaces) operators
```
nickboldt, line 10: naming question... could this be simply called "filterIIB.sh" ? does the "ForPackage" part provide more meaningful description (or does it imply there will be a different filterIIBForSomethingElse.sh in future?

product/filterIIBForPackage.sh
```diff
@@ -0,0 +1,157 @@
...
+# from an IIB image.
```
nickboldt, line 11:
```suggestion
# from an IIB image. Creates a file tree of json files, one folder per filtered package. 
```

product/copyIIBsToQuay.sh
```diff
@@ -27,7 +27,7 @@ Options:
   -p, --push                 : Push IIB(s) to quay registry; default is to show commands but not copy anything
   --force                    : If target image exists, will re-filter and re-push it; otherwise skip to avoid updating image timestamps
   -t PROD_VER                : If x.y version/tag not set, will compute from dependencies/job-config.json file
-  -o 'OCP_VER1 OCP_VER2 ...' : Space-separated list of OCP version(s) to query and publish; defaults to job-config.json values
+  -o 'OCP_VER1 OCP_VER2 ...' : Space-separated list of OCP version(s) (e.g. 'v4.10 v4.11') to query and publish; defaults to job-config.json values
```
nickboldt, line 30: can also omit the `v` prefixes...
```
    '-o') if [[ "$2" != "v"* ]]; then OCP_VERSIONS="${OCP_VERSIONS} v${2}"; else OCP_VERSIONS="${OCP_VERSIONS} ${2}"; fi; shift 1;;
```

product/copyIIBsToQuay.sh
```diff
@@ -157,33 +169,48 @@ for OCP_VER in ${OCP_VERSIONS}; do
...
+            ${FIIB} -s ${LATEST_DWO_IIB} --channel-all fast ${VERBOSEFLAG} --dir $CATALOG_DIR --packages "devworkspace-operator"
+            ${BCFF} -t ${LATEST_IIB_QUAY} --push ${VERBOSEFLAG} --dir $CATALOG_DIR --ocp-ver $OCP_VER
         else
             if [[ $VERBOSEFLAG == "-v" ]]; then echo "Copy ${LATEST_IIB_QUAY} - already exists, nothing to do"; fi
             echo "[IMG] ${LATEST_IIB_QUAY}"
         fi
         PUSHTOQUAYFORCE_LOCAL=1
     else
-        echo "${FIIB} -s ${LATEST_IIB} -t ${LATEST_IIB_QUAY} --push"
+        ${FIIB} -s ${LATEST_IIB} ${VERBOSEFLAG} --dir $CATALOG_DIR --packages "devspaces web-terminal"
```
nickboldt, line 201: missing flag ` --channel-all fast` here and on next line

product/copyIIBsToQuay.sh
```diff
@@ -157,33 +169,48 @@ for OCP_VER in ${OCP_VERSIONS}; do
...
+    if [[ $VERBOSEFLAG == "-v" ]]; then echo "Rendering catalog to $CATALOG_DIR"; fi
+
     if [[ "$PUSH" == "true" ]]; then
         # check if destination already exists in quay
         if [[ $(skopeo --insecure-policy inspect docker://${LATEST_IIB_QUAY} 2>&1) == *"Error"* ]] || [[ ${PUSHTOQUAYFORCE} -eq 1 ]]; then 
             # filter and publish to a new name, putting all operators in the fast channel
-            ${FIIB} -s ${LATEST_IIB} -t ${LATEST_IIB_QUAY} --channel-all fast --push ${VERBOSEFLAG}
+            ${FIIB} -s ${LATEST_IIB} --channel-all fast ${VERBOSEFLAG} --dir $CATALOG_DIR --packages "devspaces web-terminal"
+            ${FIIB} -s ${LATEST_DWO_IIB} --channel-all fast ${VERBOSEFLAG} --dir $CATALOG_DIR --packages "devworkspace-operator"
+            ${BCFF} -t ${LATEST_IIB_QUAY} --push ${VERBOSEFLAG} --dir $CATALOG_DIR --ocp-ver $OCP_VER
```
nickboldt, line 194: maybe we could rework this logic so the ` if [[ "$PUSH" == "true" ]]; then` check is only done for this line? that seems to be the only part that's not executed with `PUSH != true`

nickboldt, line 194:
```suggestion
            if [[ "$PUSH" == "true" ]]; then
                ${BCFF} -t ${LATEST_IIB_QUAY} --push ${VERBOSEFLAG} --dir $CATALOG_DIR --ocp-ver $OCP_VER
            else
                ${BCFF} -t ${LATEST_IIB_QUAY} ${VERBOSEFLAG} --dir $CATALOG_DIR --ocp-ver $OCP_VER
            fi
```

product/buildCatalogFromFiles.sh
amisevsk: Makes sense, I was trying to gesture at "the files had better be here when I look for them"

product/filterIIBForPackage.sh
amisevsk: Nope, just sticking with the previous naming (`filterIIBForDevSpaces`). I'll update.

product/copyIIBsToQuay.sh
amisevsk: I was fighting how to fit this in since the push and no push case also filter the IIB differently (when `--push` is specified the channel is overridden to `fast` whereas when it's not no override is done and default channel names are used.

I think it's useful to still build the image locally when push is not used. This would allow you to manually push to your own quay repo, or render out the image yourself to verify it is as expected (I used this in testing changes, for example).

Also to note, this block is already in a `if [[ "$PUSH" == "true" ]]` conditional, on line 188.

product/filterIIBForPackage.sh
amisevsk: :+1:

product/copyIIBsToQuay.sh
amisevsk: I tried it but it only added the `v` to the first one -- e.g. `-o 4.10 4.11 -> `v4.10 4.11`

product/copyIIBsToQuay.sh
amisevsk: The split was due to different behavior between push and no push -- I wasn't sure if it was intended or not so I left it as-is. I'll update this to be simpler if we always want to use `--channel-all`

nickboldt: > I think it's useful to still build the image locally when push is not used.

Yes, if you have a use case for that, then make it so.

product/copyIIBsToQuay.sh
nickboldt: well shit, that's a bug then :(

product/copyIIBsToQuay.sh
```diff
@@ -157,47 +169,57 @@ for OCP_VER in ${OCP_VERSIONS}; do
...
-        PUSHTOQUAYFORCE_LOCAL=1
+    CATALOG_DIR=$(mktemp -d --suffix "-${DS_VERSION}-${OCP_VER}-${LATEST_IIB_NUM}-$(uname -m)")
+    if [[ $VERBOSEFLAG == "-v" ]]; then echo "Rendering catalog to $CATALOG_DIR"; fi
+
+    # filter and publish to a new name, putting all operators in the fast channel
+    ${filterIIB} -s ${LATEST_IIB} --channel-all fast --dir $CATALOG_DIR --packages "devspaces web-terminal" ${VERBOSEFLAG}
+    ${filterIIB} -s ${LATEST_DWO_IIB} --channel-all fast --dir $CATALOG_DIR --packages "devworkspace-operator" ${VERBOSEFLAG}
+    if [[ "$PUSH" != "true" ]]; then
+        ${buildCatalog} -t ${LATEST_IIB_QUAY} ${VERBOSEFLAG} --dir $CATALOG_DIR --ocp-ver $OCP_VER
+        continue
```
nickboldt, line 193: what does this `continue` do?

product/copyIIBsToQuay.sh
```diff
@@ -157,47 +169,57 @@ for OCP_VER in ${OCP_VERSIONS}; do
...
+        echo "[IMG] ${LATEST_IIB_QUAY}"
     fi
+    PUSHTOQUAYFORCE_LOCAL=1
 
-    if [[ $(skopeo --insecure-policy inspect docker://${LATEST_IIB_QUAY} 2>&1) == *"Error"* ]]; then 
+    if [[ $(skopeo --insecure-policy inspect docker://${LATEST_IIB_QUAY} 2>&1) == *"Error"* ]]; then
         echo "[ERROR] Cannot find image ${LATEST_IIB_QUAY} to copy!"
         echo "[ERROR] Check output of this command for an idea of what went wrong:"
-        echo "[ERROR] ${FIIB} -s ${LATEST_IIB} -t ${LATEST_IIB_QUAY} -v --push"
+        echo "[ERROR] ${buildCatalog} -s ${LATEST_IIB} -t ${LATEST_IIB_QUAY} -v --push"
```
nickboldt, line 208: should this be:
```
echo "[ERROR] ${buildCatalog} -t ${LATEST_IIB_QUAY} --push ${VERBOSEFLAG} --dir $CATALOG_DIR --ocp-ver $OCP_VER"
```
?

product/copyIIBsToQuay.sh
amisevsk: Yup, the rest of the steps require push -- mostly they're skopeo copy, etc., which relies on the image being pushed already. I'll update with a comment explaining this.

product/copyIIBsToQuay.sh
amisevsk: Yup! Docs are always the last thing to be updated :)

nickboldt (approved):

PR #835, opened 2022-10-31
Set `code` to be built from `7.56.x` (or fall back to 7.55.x if no 7.56.x branch exists yet)

azatsarynnyy:
### What does this PR do?
Sets `code` to be built from `7.56.x` (or fall back to 7.55.x if no 7.56.x branch exists yet)

### DIFF
--- a/dependencies/job-config.json
+++ b/dependencies/job-config.json
@@ -28,8 +28,8 @@
       },
       "3.3": {
         "upstream_branch": [
-          "devspaces-3.3-rhel-8",
-          "devspaces-3.3-rhel-8"
+          "7.54.x",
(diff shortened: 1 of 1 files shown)


nickboldt:
3.3 will be based on 7.56 (not 7.54) so I've updated the PR.

nickboldt:
Redundant change to this fix to the automation https://github.com/redhat-developer/devspaces/commit/657f2f5a64a2197d30b33a0881b6b4111481a7da and this run of the script to change and commit the change: https://github.com/redhat-developer/devspaces/commit/f98a171b40400a290181a6d275b31d8e23c4a9d7

PR #837, opened 2022-11-03
CRW-3138 proposed change to an operator dependency name.

SDawley:
### What does this PR do?
I just want to change this one name to better fit how it will be used in devspaces-operators container.yaml

### DIFF
--- a/dependencies/job-config.json
+++ b/dependencies/job-config.json
@@ -1007,7 +1007,7 @@
       "3.3": "next",
       "3.x": "next"
     },
-    "DEV_WORKSPACE_CONTROLLER_VERSION": {
+    "DEV_WORKSPACE_CONTROLLER": {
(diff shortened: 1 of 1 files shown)


dependencies/job-config.json
```diff
@@ -1007,7 +1007,7 @@
       "3.3": "next",
       "3.x": "next"
     },
-    "DEV_WORKSPACE_CONTROLLER_VERSION": {
+    "DEV_WORKSPACE_CONTROLLER": {
```
nickboldt, line 1010: I'm OK with this rename as long as ALL places that refer to it are not broken as a result... eg., might have to backport part of your changes to the 3.2 branch too in https://github.com/redhat-developer/devspaces-images/pull/344#pullrequestreview-1167249630 just in case we need to rebuild a 3.2.1 operator.

PR #863, opened 2022-11-28
Adding required registries to OCP image configuration when restricted

oybed:
### What does this PR do?
This PR updates the `installDevSpacesFromLatestIIB` script to enable required registries in the `allowedRegistries` / `allowedRegistriesForImport` cluster configuration when the cluster is already restricted. The script will not add these if no restriction exists.

### DIFF
--- a/product/installDevSpacesFromLatestIIB.sh
+++ b/product/installDevSpacesFromLatestIIB.sh
@@ -49,6 +49,14 @@ CHANNEL_DS="stable"
 # if you want your own registry here, use --icsp flag to specify it
 ICSP_FLAGs=""
 
+# Additional registries that needs to be added to the OpenShift configuration
+# in the event that the OpenShift cluster is restricted to not allow these registries
+ADDITIONAL_REGISTRIES=(
+  "registry-proxy.engineering.redhat.com"
+  "registry.stage.redhat.io"
+  "quay.io"
+)
+
 errorf() {
   echo -e "${RED}Error: $1${NC}"
 }
@@ -256,6 +264,8 @@ done
 
 preflight
 
+TMPDIR=$(mktemp -d)
+
 # detect openshift server version and arch; for amd, use x86_64; for others, trim linux/ prefix
 OPENSHIFT_VER=$(oc version -o json | jq -r '.openshiftVersion | scan("^[0-9].[0-9]+")')
 OPENSHIFT_ARCH=$(oc version -o json | jq -r '.serverVersion.platform' | sed -r -e "s#linux/amd64#x86_64#" -e "s#linux/##")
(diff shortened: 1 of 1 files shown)


oybed:
CC @l0rd - per our convo last week

nickboldt:
Is this something you could rebase against the current version?

oybed:
@nickboldt rebase done ... I won't be able to re-test, however.

product/installDevSpacesFromLatestIIB.sh
```diff
@@ -49,6 +49,14 @@ CHANNEL_DS="stable"
 # if you want your own registry here, use --icsp flag to specify it
 ICSP_FLAGs=""
 
+# Additional registries that needs to be added to the OpenShift configuration
+# in the event that the OpenShift cluster is restricted to not allow these registries
+ADDITIONAL_REGISTRIES=(
+  "registry-proxy.engineering.redhat.com"
+  "registry.stage.redhat.io"
```
nickboldt, line 56: you don't need registry.redhat.io or registry.access.redhat.com in here?

product/installDevSpacesFromLatestIIB.sh
```diff
@@ -290,6 +300,40 @@ if [[ ${ENABLE_CATALOGSOURCES} == "true" ]]; then
...
+      imageRegistries=$(echo -n ${imageRegistries} | jq ".spec.allowedRegistriesForImport += [{\"domainName\": \"${r}\",\"insecure\":false}]")
+    fi
+  done
+
+  # Remove a few entries in the JSON object to allow for oc apply below
+  ImagesJson=$(echo -n ${imageRegistries} | jq 'del(.metadata.creationTimestamp, .metadata.generation, .metadata.ownerReferences, .metadata.resourceVersion, .metadata.uid)')
+
+  echo -n "${imageRegistries}" | oc apply -f -
+
+  echo "[INFO] OpenShift Cluster image registries updated."
```
nickboldt, line 332: should the list of registries added be echoed here from `ADDITIONAL_REGISTRIES` ?

product/installDevSpacesFromLatestIIB.sh
```diff
@@ -290,6 +300,40 @@ if [[ ${ENABLE_CATALOGSOURCES} == "true" ]]; then
...
+# Function used to check to see if additional registries are needed
+# in the allowedRegistries / allowedRegistriesForImport cluster configuration 
+# - required registries will be added if missing
+# - registry list will be updated *only* if Cluster Configuration is set to a restricted list
+addRequiredRegistries() {
+  imageRegistries=$(oc get image.config.openshift.io/cluster -o json)
+
+  # Check to see if the Cluster Configuration is restricted to a specific list of registries
+  if [[ $(echo -n ${imageRegistries} | jq '.spec.registrySources.allowedRegistries | length') -eq 0 ]]; then
+    echo "[INFO] OpenShift environment does not have restricted registries - continuing..."
```
nickboldt, line 312:
```suggestion
    echo "[INFO] OpenShift environment does not have restricted registries - nothing to add."
```

product/installDevSpacesFromLatestIIB.sh
oybed: Sounds like a good addition - will add

product/installDevSpacesFromLatestIIB.sh
nickboldt: I'm curious, have you used the latest `dsc` binary to do an install? 

* https://developers.redhat.com/products/openshift-dev-spaces/download (official releases)
* https://github.com/redhat-developer/devspaces-chectl/releases/ (pre-releases)

It now supports installation from `:latest` and `:next` builds from https://quay.io/repository/devspaces/iib?tab=tags using the `fast` channel subscription. 

See readme at https://github.com/redhat-developer/devspaces-chectl/#dsc-serverdeploy for full command usage, but here's the gist:
```
dsc server:deploy --olm-channel={stable|latest|fast|next}
```

oybed: Our work with the customer wrapped up in December, so have not re-tried any of this since then. We were guided to use this bash script back then to work with pre-released versions. 

This PR was submitted to support a missing piece of the bash script that caused it not to work for us (and we wasted quite a bit of time troubleshooting this), hence was hoping this would be helpful for others to avoid a similar experience.

tolusha: @nickboldt 
I don't modify `image.config.openshift.io/cluster` resource, so changes deserve to be included into `dsc`  as well

nickboldt:
Moved to https://issues.redhat.com/browse/CRW-4180 to be triaged in a future sprint.

nickboldt:
> I won't be able to re-test, however.

I could merge this as is, but it would be nice to know it's working as expected w/ the latest version; or we could wait and just implement it for `dsc`. @oybed I'm guessing since you're not using this currently, there's no urgency here, right?

nickboldt:
I'm going to close this until it can be tested; meanwhile we can look at implementing similar feature in dsc in the future.

PR #869, opened 2022-12-27
chore: add container-contribution attribute into the editor definition

svor:

### What does this PR do?
Adds container-contribution attribute for VS Code and Theia editors.

**_NOTE_: The IDEA editor wasn't changed**

https://issues.redhat.com/browse/CRW-3612


nickboldt:
+1 but should we be adding features to theia? we could do this for code-rhel8 only, as theia is deprecated.

svor:
The mechanism of DevWorkspace generation was reworked https://github.com/redhat-developer/devspaces/pull/870
Since we still generate DevWorkspace with che-theia editor, I'd proceed with current changes

PR #877, opened 2023-01-25
Add rule about "need triage by PM, CEE” issue label''; fix typo.

dmytro-ndp:
### What does this PR do?
Improve issues-and-triage.adoc

### DIFF
--- a/issues-and-triage.adoc
+++ b/issues-and-triage.adoc
@@ -50,7 +50,7 @@ For some issues, you may be able to add even more information:
 . Set **Assignee** if not set and work will begin immediately; remove assignee if work is not scheduled
 . Set **Fix Version** if known/obvious; empty and "3.x" versions can be set to a more concrete value during sprint planning
 . Set **Priority** if default “major” is not correct. 
-Don't change priority if the issue has a label `need triage by PM/Support`, but you are not Dev Spaces / DWO / WTO Product Manager or CEE.
+Don't change priority if the issue has a label `need-PM/Support-triage`, but you are not Dev Spaces / DWO / WTO Product Manager or CEE.
 
 . Set **release note fields** if relevant:
     - Affects: Release Notes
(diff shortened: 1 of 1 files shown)


RickJWagner:
Please note that the support organization ('CEE') has been reorganized.  What was formerly called 'CEE' is now called 'XE', though a part of it still is called CEE.
Probably best to call it 'Support', because org names have shifted several times, will probably continue to do so.

issues-and-triage.adoc
```diff
@@ -49,7 +49,9 @@ For some issues, you may be able to add even more information:
 
 . Set **Assignee** if not set and work will begin immediately; remove assignee if work is not scheduled
 . Set **Fix Version** if known/obvious; empty and "3.x" versions can be set to a more concrete value during sprint planning
-. Set **Priority** if default “major” is not correct
+. Set **Priority** if default “major” is not correct. 
+Don't change priority if the issue has a label `need-PM/Support-triage`, but you are not Dev Spaces / DWO / WTO Product Manager or CEE.
```
RickJWagner, line 53: Please change 'CEE' to 'Support'.
CEE (which used to be called 'GSS', etc.) has been re-organized again, it is now called 'XE' but probably will be something different again in the near future.  Calling it 'Support' is probably safest for the long run.

nickboldt: This change works for me, but bearing in mind that PM is only 1/10th of the triage rotation and CEE isn't in there at all. https://github.com/eclipse/che/wiki/Issue-Tracking#triage-curators

So either we need to add Rick into the rotation, or we need to include a "hey rick, please review these issues" step in the process when there are fresh issues w/ the new label.

nickboldt:
I've merged your suggestion @dmytro-ndp https://github.com/redhat-developer/devspaces/commit/397039d0ee3eee6ff5e67490112888b7e98c3253  with some tweaks: 

https://github.com/redhat-developer/devspaces/commit/2b1687500940cd51b84d9873ac9027671189407f

PR #893, opened 2023-02-10
feat: add latest floating tag for DevSpace Samples during release tagging process

mkuznyetsov:

### What does this PR do?
For devspaces samples, a floating tag 'latest' will exist that will point to the latest release version of the sample
When tagging release, samples projects will have the 'latest' tagged updated with the release version (except for the cases, when a "lower" bugfix version like 3.5.1 gets released, when 3.6.0 already exists)

https://issues.redhat.com/browse/CRW-3805

### DIFF
--- a/product/tagRelease.sh
+++ b/product/tagRelease.sh
@@ -197,14 +197,14 @@ pushBranchAndOrTagGH () {
 		git push origin "${CSV_VERSION}" || true
 		# update latest floating tag for samples
 		if [[ $org == "${samplesRepo}" ]]; then
-			# check all 3.*.* tags in the repo, if tag exists with a higher version than the new tag
+			# check all of the 3.*.* tags in the repo, if tag exists with a higher version than the new tag
 			# then latest floating tag should not be updated
 			LATEST_TAG=$(git tag -l -n 3.*.* --sort -version:refname | head -n 1 | grep "3.*.*")
 			if { echo "${CSV_VERSION}"; echo "${LATEST_TAG}"; } | sort --version-sort --check;then
-				echo "[DEBUG] Sample version ${d} is less than latest version";
+				echo "[DEBUG] Sample version ${d} is less than latest ${LATEST_TAG} version, not updating the latest tag";
 			else
 				echo "[DEBUG] Updating latest tag of sample ${d} with ${CSV_VERSION}"
-				git tag -D "latest"
+				git tag -d "latest"
 				git tag "latest"
 				git push origin "latest" -f || true
 			fi
--- a/product/tagRelease.sh
+++ b/product/tagRelease.sh
@@ -195,10 +195,20 @@ pushBranchAndOrTagGH () {
 	if [[ $CSV_VERSION ]]; then # push a new tag (or no-op if exists)
 		git tag "${CSV_VERSION}" || true
 		git push origin "${CSV_VERSION}" || true
-    # for sample projects, update (force push) the update
-    if [[ $org == "${samplesRepo}" ]]; then
-      git push origin "${CSV_VERSION}:latest" -f || true
-    fi
+		# update latest floating tag for samples
+		if [[ $org == "${samplesRepo}" ]]; then
+			# check all 3.*.* tags in the repo, if tag exists with a higher version than the new tag
+			# then latest floating tag should not be updated
+			LATEST_TAG=$(git tag -l -n 3.*.* --sort -version:refname | head -n 1 | grep "3.*.*")
+			if { echo "${CSV_VERSION}"; echo "${LATEST_TAG}"; } | sort --version-sort --check;then
+				echo "[DEBUG] Sample version  is less than latest version";
+			else
+				echo "[DEBUG] Updating latest tag of sample $d with ${CSV_VERSION}"
+				git tag -D "latest"
+				git tag "latest"
+				git push origin "latest" -f || true
+			fi
+		fi
 	fi
 	popd >/dev/null || exit 1
 }
--- a/product/tagRelease.sh
+++ b/product/tagRelease.sh
@@ -199,14 +199,14 @@ pushBranchAndOrTagGH () {
 		if [[ $org == "${samplesRepo}" ]]; then
 			# check all of the 3.*.* tags in the repo, if tag exists with a higher version than the new tag
 			# then latest floating tag should not be updated
-			LATEST_TAG=$(git tag -l -n 3.*.* --sort -version:refname | head -n 1 | grep "3.*.*")
+			LATEST_TAG=$(git tag -l -n 3.*.* --sort -version:refname | head -n 1 | grep -o "3.[0-9]*.[0-9]*")
 			if { echo "${CSV_VERSION}"; echo "${LATEST_TAG}"; } | sort --version-sort --check;then
 				echo "[DEBUG] Sample version ${d} is less than latest ${LATEST_TAG} version, not updating the latest tag";
 			else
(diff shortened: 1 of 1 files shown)


product/tagRelease.sh
```diff
@@ -195,6 +195,10 @@ pushBranchAndOrTagGH () {
 	if [[ $CSV_VERSION ]]; then # push a new tag (or no-op if exists)
 		git tag "${CSV_VERSION}" || true
 		git push origin "${CSV_VERSION}" || true
+    # for sample projects, update (force push) the update
+    if [[ $org == "${samplesRepo}" ]]; then
+      git push origin "${CSV_VERSION}:latest" -f || true
```
nickboldt, line 200: this will create a branch called `3.6:latest` which is not what we want. 

The goal is to have a branch that always contains the `latest`.

product/tagRelease.sh
```diff
@@ -195,6 +195,10 @@ pushBranchAndOrTagGH () {
 	if [[ $CSV_VERSION ]]; then # push a new tag (or no-op if exists)
 		git tag "${CSV_VERSION}" || true
 		git push origin "${CSV_VERSION}" || true
+    # for sample projects, update (force push) the update
```
nickboldt, line 198: once again, your indents don't line up with anything. Please install a linter.

product/tagRelease.sh
```diff
@@ -195,6 +195,10 @@ pushBranchAndOrTagGH () {
...
+    if [[ $org == "${samplesRepo}" ]]; then
```
nickboldt, line 199: maybe you can move this up into like 188-190 ? Seems like updating the **branch** should happend the same time as a push to **TARGET_BRANCH** rather than when we're pushing **tags**.

product/tagRelease.sh
nickboldt: occurs to me a **tag** called `latest` is probably better than a branch, and we should (as you suggested here) create it the same time we're tagging a 3.yy release.

nickboldt, line 200:
```suggestion
            git tag "latest" -D || true
            git tag "latest"
            git push origin "latest" -f || true
```

mkuznyetsov: just occured to me that tag alone is probably good enough too, if it's for the usecase given in the issue

but if anything, that semicolon is a refspec thing, that meant to that local branch `CSV_VERSION` (or rather `TARGET_BRANCH` would be pushed as `latest` on remote

nickboldt: why wouldn't a tag work? whole point is to have a stable URL to embed in https://www.eclipse.org/che/getting-started/ instead of a URL that changes with every DS release.

product/tagRelease.sh
```diff
@@ -195,6 +195,20 @@ pushBranchAndOrTagGH () {
...
+		# update latest floating tag for samples
+		if [[ $org == "${samplesRepo}" ]]; then
+			# check all of the 3.*.* tags in the repo, if tag exists with a higher version than the new tag
+			# then latest floating tag should not be updated
+			LATEST_TAG=$(git tag -l -n 3.*.* --sort -version:refname | head -n 1 | grep "3.*.*")
+			if { echo "${CSV_VERSION}"; echo "${LATEST_TAG}"; } | sort --version-sort --check;then
+				echo "[DEBUG] Sample version ${d} is less than latest ${LATEST_TAG} version, not updating the latest tag";
+			else
+				echo "[DEBUG] Updating latest tag of sample ${d} with ${CSV_VERSION}"
+				git tag -d "latest"
```
nickboldt, line 207: this could fail if the tag doesn't exist? Do we need an `||true` ?

product/tagRelease.sh
```diff
@@ -195,6 +195,20 @@ pushBranchAndOrTagGH () {
...
+				git tag "latest"
+				git push origin "latest" -f || true
```
nickboldt, line 209: should this be allowed to fail? surely this should always pass...

product/tagRelease.sh
```diff
@@ -195,6 +195,20 @@ pushBranchAndOrTagGH () {
 	if [[ $CSV_VERSION ]]; then # push a new tag (or no-op if exists)
 		git tag "${CSV_VERSION}" || true
 		git push origin "${CSV_VERSION}" || true
+		# update latest floating tag for samples
+		if [[ $org == "${samplesRepo}" ]]; then
+			# check all of the 3.*.* tags in the repo, if tag exists with a higher version than the new tag
+			# then latest floating tag should not be updated
+			LATEST_TAG=$(git tag -l -n 3.*.* --sort -version:refname | head -n 1 | grep "3.*.*")
+			if { echo "${CSV_VERSION}"; echo "${LATEST_TAG}"; } | sort --version-sort --check;then
```
nickboldt, line 203: interesting approach. LATEST_TAG might include extra crap like
`3.4.0           chore(devfile) link v2 devfile to :3.4; set image refs to registry.redhat.io/devspaces/`

(looking at https://github.com/devspaces-samples/lombok-project-sample)

but your test seems to work.

What happens with a repo where there's no previous tag? Also seems to work! 

(looking at https://github.com/devspaces-samples/php-hello-world/tree/devspaces-3-rhel-8 which has no tags)

product/tagRelease.sh
```diff
@@ -195,6 +195,20 @@ pushBranchAndOrTagGH () {
...
+				echo "[DEBUG] Sample version ${d} is less than latest ${LATEST_TAG} version, not updating the latest tag";
```
nickboldt, line 204: this message will look weird with the extra stuff noted above...

```
[DEBUG] Sample version lombok-project-sample is less than latest 3.4.0           chore(devfile) link v2 devfile to :3.4; set image refs to registry.redhat.io/devspaces/ version, not updating the latest tag
```
or 
```
[DEBUG] Sample version php-hello-world is less than latest  version, not updating the latest tag
```

product/tagRelease.sh
mkuznyetsov: yes

product/tagRelease.sh
mkuznyetsov: thx, adjusted the grep at the end to trim right after 3.x.y

product/tagRelease.sh
mkuznyetsov: should be good now

product/tagRelease.sh
nickboldt: nice ! didn't know about the `-o` flag

product/tagRelease.sh
```diff
@@ -199,14 +199,14 @@ pushBranchAndOrTagGH () {
...
+			LATEST_TAG=$(git tag -l -n 3.*.* --sort -version:refname | head -n 1 | grep -o "3.[0-9]*.[0-9]*")
 			if { echo "${CSV_VERSION}"; echo "${LATEST_TAG}"; } | sort --version-sort --check;then
 				echo "[DEBUG] Sample version ${d} is less than latest ${LATEST_TAG} version, not updating the latest tag";
 			else
 				echo "[DEBUG] Updating latest tag of sample ${d} with ${CSV_VERSION}"
-				git tag -d "latest"
+				git tag -d "latest" || true
 				git tag "latest"
-				git push origin "latest" -f || true
+				git push origin "latest"
```
nickboldt, line 209: you removed the force. Why?

mkuznyetsov: right, force should be there after all

nickboldt:
Merged! Can you run the script and verify new `latest` tags are created when you run it, as if you were tagging for the 3.4.0 release ? Can you also verify they're recreated if you run the script again, and that nothing destructive happens?

PR #899, opened 2023-02-17
fix: fixes to tagRelease.sh for tagging samples

mkuznyetsov:
### What does this PR do?
1. on tagging operation, if PD branch is not sent, then tagging of PD projects will not happen (the condition for it is already there, but pd branch would have a default target branch), could help to save time, if this script needs to be rerun for non-pd projects?
2. Improve latest tag selection condition, to avoid false matches in some usecase


nickboldt:
1. I don't understand your first change. Just pass in a pkgs.devel value if you don't like the default.

2. In which case will there be a tag that contains 3.<something> but doesn't START with 3.<something> ?

mkuznyetsov:
1. Alright, I'll remove it
2. grep matchers the entire string, which contains tag description

PR #920, opened 2023-03-22
fix: Do not duplicate che-theia commands

RomanNikitenko:
### What does this PR do?
Rename `preStart` che-theia command.
We need this change because of upstream changes: https://github.com/eclipse-che/che-plugin-registry/pull/1613

Related to https://github.com/eclipse/che/issues/22071


svor (approved):
LGTM

PR #944, opened 2023-04-25
Set "CSV_VERSION_PREV": "3.5.0-0.1679521107.p"

dmytro-ndp:
### What does this PR do?
It aligns Dev Spaces 3.6.0 to recent version of Dev Spaces 3.5.0-0.1679521107.p, and 3.7.0 to 3.6.0.

https://issues.redhat.com/browse/CRW-4324

### DIFF
--- a/dependencies/job-config.json
+++ b/dependencies/job-config.json
@@ -531,7 +531,7 @@
       },
       "3.x": {
         "CSV_VERSION": "3.7.0",
-        "CSV_VERSION_PREV": "3.6.0"
+        "CSV_VERSION_PREV": "3.5.0-0.1679521107.p"
       }
     }
   },
--- a/dependencies/job-config.json
+++ b/dependencies/job-config.json
@@ -523,7 +523,7 @@
       },
       "3.6": {
         "CSV_VERSION": "3.6.0",
-        "CSV_VERSION_PREV": "3.5.0-0.1679521107.p""
+        "CSV_VERSION_PREV": "3.5.0-0.1679521107.p"
       },
       "3.7": {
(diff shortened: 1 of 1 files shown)


dependencies/job-config.json
```diff
@@ -523,15 +523,15 @@
...
       },
       "3.7": {
         "CSV_VERSION": "3.7.0",
-        "CSV_VERSION_PREV": "3.5.0"
+        "CSV_VERSION_PREV": "3.6.0"
       },
       "3.x": {
         "CSV_VERSION": "3.7.0",
-        "CSV_VERSION_PREV": "3.5.0"
+        "CSV_VERSION_PREV": "3.6.0"
```
nickboldt, line 534: There is no 3.6.0, so this should be the last released version

```suggestion
        "CSV_VERSION_PREV": "3.5.0-0.1679521107.p"
```

nickboldt:
Unless you plan to do testing of 3.5 -> 3.6RC -> 3.7 CI upgrade path, the correct `_PREV` value for 3.7 and 3.x is the latest 3.5.0.fm.

Otherwise you'll get errors trying to install 3.7 as there is no released 3.6 yet to resolve the upgrade tree.

dmytro-ndp:
If I were asked to test an upgrade to 3.7 CI, I would test the 3.5 -> 3.6RC -> 3.7 CI upgrade path as a more relevant option.

nickboldt:
If we set 3.7 as a `replaces: 3.6` then you won't be able to install 3.7 without first installing 3.6.

we had this problem 6 weeks ago, and 12 weeks ago...

PR #947, opened 2023-04-26
feat:bump devworkspace generator tool

svor:
### What does this PR do?
bump evworkspace generator tool from 0.0.1-c8bd5c0 to 0.0.1-638fadc

https://github.com/eclipse/che/issues/22149


nickboldt:
two questions:
* does this need a backport in 3.6?
* once we have plugin and devfile reg using cachito, I'm assuming we won't need to hardcode anything and will just use latest SHA from the appropriate branch

PR #965, opened 2023-06-09
chore: add devfile.yaml to work with plugin registry

svor:
### What does this PR do?
Introduces devfile.yaml with commands that allow to verify changes made in the plugin registry.

It is important to have an access to modify CheCluster definition on OS cluster

Only one extension in the list of the embedded plugin registry:
screenshot-nimbusweb me-2023 06 09-16_30_25

Rresult of the changes:
screenshot-nimbusweb me-2023 06 09-16_32_31

Related issue: https://github.com/eclipse/che/issues/22243

### DIFF
--- a/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
+++ b/dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
@@ -47,7 +47,7 @@ WORKDIR /
 
 # Copy OpenVSX server files
 COPY --chown=0:0 /openvsx-server.tar.gz .
-RUN tar -xf openvsx-server.tar.gz && rm openvsx-server.tar.gz
+RUN tar --no-same-owner -xf openvsx-server.tar.gz && rm openvsx-server.tar.gz
 # Copy our configuration file for OpenVSX server
 COPY /build/dockerfiles/application.yaml /openvsx-server/config/
 RUN chmod -R g+rwx /openvsx-server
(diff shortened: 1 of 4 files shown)


dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -47,7 +47,7 @@ WORKDIR /
 
 # Copy OpenVSX server files
 COPY --chown=0:0 /openvsx-server.tar.gz .
-RUN tar -xf openvsx-server.tar.gz && rm openvsx-server.tar.gz
+RUN tar --no-same-owner -xf openvsx-server.tar.gz && rm openvsx-server.tar.gz
```
svor, line 50: to avoid errors when building on dogfooding instance
```
tar: ./output/v3/plugins/che-incubator/che-code/latest/devfile.yaml: Cannot change ownership to uid 1001290000, gid 1001290000: Invalid argument
./output/v3/plugins/che-incubator/che-code/insiders/
```

nickboldt: maybe you prefer the equivalent of `--chown=0:0` to make the files owned by root instead of the non-root openshift userid?

dependencies/che-plugin-registry/build/dockerfiles/Dockerfile
```diff
@@ -47,7 +47,7 @@ WORKDIR /
```
nickboldt, line 47: you set `WORKDIR /` twice on lines 17 and 46.

PR #968, opened 2023-06-14
Add arguments to download_vsix.sh 

SDawley:
Add arguments to download_vsix.sh so it can be run to update openvsx-sync.json for building internal plugins.

https://issues.redhat.com/browse/CRW-3336

Added option arguments to download_vsix.sh. If no arguments are provided it should run as it originally did, but can now also be used to generate an updated version of openvsx-sync.json that can be used by the pluginregistry-plugins build to build vsix versions internally.

Goal is to use the same version determination process so that the versions in Che and DevSpaces stay as close as possible.

### DIFF
--- a/dependencies/che-plugin-registry/build/scripts/download_vsix.sh
+++ b/dependencies/che-plugin-registry/build/scripts/download_vsix.sh
@@ -3,16 +3,45 @@
 set -e
 set -o pipefail
 
-scriptsBranch="$(git rev-parse --abbrev-ref HEAD 2>/dev/null || true)"
-if [[ $scriptsBranch != "devspaces-3."*"-rhel-8" ]]; then
-    scriptsBranch="devspaces-3-rhel-8"
+downloadVsix=1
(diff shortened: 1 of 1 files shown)


dependencies/che-plugin-registry/build/scripts/download_vsix.sh
```diff
@@ -3,16 +3,45 @@
...
+  case $1 in
+    '-b'|'--branch') scriptsBranch="$2"; shift 1;;
+    '-j'|'--json') openvsxJson="$2"; shift 1;;
+    '--no-download') downloadVsix=0;;
+    '-h'|'--help') usage;;
+  esac
+  shift 1
+done
+
+if [[ ! "${scriptsbranch}" ]]; then
```
svor, line 32:
```suggestion
if [[ ! "${scriptsBranch}" ]]; then 
```

svor:
@SDawley please fix shellcheck errors

svor (approved):
I've haven't tested it but the codebase looks good

PR #969, opened 2023-06-15
Avoid overwriting `JAVA_HOME` by IDEA

azatsarynnyy:
### What does this PR do?
In the UDI image, `/home/user/.java` is a `JAVA_HOME`.
IDEA editor is overwriting that directory by its config.
This patch changes the IDEA config path.

https://issues.redhat.com/browse/CRW-4525

### DIFF
--- a/dependencies/che-plugin-registry/che-editors.yaml
+++ b/dependencies/che-plugin-registry/che-editors.yaml
@@ -41,7 +41,7 @@ editors:
             - name: projector-configuration
               path: /home/user/.jetbrains
             - name: projector-java-configuration
-              path: /home/user/.java
+              path: /home/user/.projector-java
           memoryLimit: 6144Mi
           memoryRequest: 2048Mi
           cpuLimit: 2000m


l0rd (approved):

l0rd:
@azatsarynnyy shouldn't you update JetBrains IDEs config to reflect that now projector Java configuration is in `${HOME}/.projector-java`?

azatsarynnyy:
> @azatsarynnyy shouldn't you update JetBrains IDEs config to reflect that now projector Java configuration is in `${HOME}/.projector-java`?

@l0rd Sorry, I didn't get your idea about updating the JetBrains IDEs config. Do you mean this environment variable?
```yaml
- name: PROJECTOR_CONFIG_DIR
  value: /home/user/.jetbrains
```

Why do you think we need to change it?

l0rd:
What I mean is that projector is currently using the folder `~/.java` to store its configuration. The fact that you have changed the folder of the volume doesn't affect projector configuration: `~/.projector-java` will be persisted but I suspect that projector will continue to write its configuration in `~/.java`.

azatsarynnyy:
Yes, we need to tell our Jetbrains editors to use another folder.
I checked why do we need a volume for it:
JetBrains IDEs store some values to java preferences. Like [accepted EULA](https://github.com/eclipse/che/issues/21116).
On Linux, Java Preferences API storage is in the `$HOME/.java/.userPrefs/prefs.xml` directory, by default.
Having a volume for `/home/user/.java` allows restoring such java preferences for IDEA across workspace restarts.

The only idea I have for now is passing something like `-Djava.util.prefs.systemRoot=/home/user/.projector-java` when running IDE. But it seems too fragile for me.

l0rd:
> The only idea I have for now is passing something like -Djava.util.prefs.systemRoot=/home/user/.projector-java when running IDE. But it seems too fragile for me.

Why do you think it's fragile? What could go wrong?

Also another alternative is to copy `$HOME/.java/.userPrefs/prefs.xml` to `$HOME/.jetbrains` at container startup and restore it at startup...

PR #974, opened 2023-07-05
Update getLatestImageTags.sh

SDawley:
### What does this PR do?
Fix regex pattern for Freshmaker build numbers to find builds with more than one base digit.
The asterisk is very powerful...

Lumping this in with https://issues.redhat.com/browse/CRW-4577


nickboldt:
ncie catch!

PR #975, opened 2023-07-06
check vscode extensions' compatibility

svor:
### What does this PR do?
Add a script to validate the compatibility between the engine versions of VS Code extensions (downloaded for the embedded plugin registry) and a Che Code editor version. 

It ensures that the extensions' engine versions are not ahead of the code version to avoid compatibility issues.

This script executes during the build and it will fail if any of the downloaded extensions is not compatible with current version of che code editor. 

To check the execution, take a look at the PR check GH action https://github.com/redhat-developer/devspaces/actions/runs/5475611769/jobs/9971952987?pr=975#step:8:1159

https://issues.redhat.com/browse/CRW-4115

#### Release Notes

N/A

#### Docs PR (if applicable)

N/A


dependencies/che-plugin-registry/build/scripts/test/check_compatibility.sh
```diff
@@ -0,0 +1,76 @@
...
+  vscodeEngineVersion="${vscodeEngineVersion//^/}"
+  # remove >= from the engine version
+  vscodeEngineVersion="${vscodeEngineVersion//>=/}"
+  # replace x by 0 in the engine version
+  vscodeEngineVersion="${vscodeEngineVersion//x/0}"
+  # check if the extension's engine version is compatible with the code version
+  # if the extension's engine version is ahead of the code version, check a next version of the extension
+  if [[  "$vscodeEngineVersion" = "$(echo -e "$vscodeEngineVersion\n$codeVersion" | sort -V | head -n1)" ]]; then
+    #VS Code version >= Engine version, can proceed."
+    echo -e "${GREEN}${EMOJI_PASS}${RESETSTYLE} The engine versin is compatible in $archive!"
```
nickboldt, line 70:
```suggestion
    echo -e "${GREEN}${EMOJI_PASS}${RESETSTYLE} Engine version compatible with $archive!"
```

dependencies/che-plugin-registry/build/scripts/test/check_compatibility.sh
```diff
@@ -0,0 +1,76 @@
...
+BOLD="\e[1m"
+DEFAULT_EMOJI_HEADER="🏃" # could be overiden with EMOJI_HEADER="-"
+EMOJI_HEADER=${EMOJI_HEADER:-$DEFAULT_EMOJI_HEADER}
+DEFAULT_EMOJI_PASS="✔" # could be overriden with EMOJI_PASS="[PASS]"
+EMOJI_PASS=${EMOJI_PASS:-$DEFAULT_EMOJI_PASS}
+DEFAULT_EMOJI_FAIL="✘" # could be overriden with EMOJI_FAIL="[FAIL]"
+EMOJI_FAIL=${EMOJI_FAIL:-$DEFAULT_EMOJI_FAIL}
+
+function initTest() {
+  echo -e "${BOLD}\n${EMOJI_HEADER} ${1}${RESETSTYLE}"
```
nickboldt, line 34:
```suggestion
  echo -n -e "${BOLD}\n${EMOJI_HEADER} ${1}${RESETSTYLE} ... "
```

dependencies/che-plugin-registry/build/scripts/test/check_compatibility.sh
```diff
@@ -0,0 +1,76 @@
...
+  vscodeEngineVersion="${vscodeEngineVersion//>=/}"
+  # replace x by 0 in the engine version
+  vscodeEngineVersion="${vscodeEngineVersion//x/0}"
+  # check if the extension's engine version is compatible with the code version
+  # if the extension's engine version is ahead of the code version, check a next version of the extension
+  if [[  "$vscodeEngineVersion" = "$(echo -e "$vscodeEngineVersion\n$codeVersion" | sort -V | head -n1)" ]]; then
+    #VS Code version >= Engine version, can proceed."
+    echo -e "${GREEN}${EMOJI_PASS}${RESETSTYLE} The engine versin is compatible in $archive!"
+  else 
+    echo -e "Extension's engine version is ahead of Che Code version ($codeVersion)."
```
nickboldt, line 72:
```suggestion
    echo -e "Extension requires a newer engine version than Che Code version ($codeVersion)."
```

nickboldt (approved):
would be nice to see log results when a failure occurs, but +1 from me.

dependencies/che-plugin-registry/build/scripts/test/check_compatibility.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+# available under the terms of the Eclipse Public License 2.0
```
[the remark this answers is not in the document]
svor: fixed

svor:
> ```
> - Should replace image references in external_images.txt with RELATED_IMAGE env vars
> --------------------------Digest map--------------------------
> Key: registry.redhat.io/devspaces/udi-rhel8:2.16 Value: @sha256:becfa80ae0e0e86f815e8981c071a68952b6a488298d75257515855[38](https://github.com/redhat-developer/devspaces/actions/runs/5475611769/jobs/9971952987?pr=975#step:8:39)a14d88e
> --------------------------------------------------------------
> [PASS] Test passed!
> 
> - Should replace image references in che-code devfile.yaml with RELATED_IMAGE env vars 
> --------------------------Digest map--------------------------
> Key: registry.redhat.io/devspaces/udi-rhel8:2.16 Value: @sha256:becfa80ae0e0e86f815e8981c071a68952b6a488298d7525751585538a14d88e
> Key: registry.redhat.io/devspaces/udi-rhel8:3.5 Value: @sha256:99ff1b5c541855e4cf368816c4bcdcdc86d3230[40](https://github.com/redhat-developer/devspaces/actions/runs/5475611769/jobs/9971952987?pr=975#step:8:41)23f72c4443213a4032ef05b
> Key: registry.redhat.io/devspaces/code-rhel8:3.5 Value: @sha256:debc18de31a6b3b575e42cc485f6c22[41](https://github.com/redhat-developer/devspaces/actions/runs/5475611769/jobs/9971952987?pr=975#step:8:42)ee4d3d6988fad4e4e9837edba24f89f
> --------------------------------------------------------------
> [PASS] Test passed!
> Build with docker build
> Error: No such image: che-ovsx:latest
> Error: No such image: che-openvsx:latest
> Sending build context to Docker daemon  155.6kB
> ```
> 
> Is that error normal and if so, should it be downgraded to a warning?

I think the error is normal, but I'm not sure if it's possible to change the status. This message happens when the builder is trying to cleanup images from local cache https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/dependencies/che-plugin-registry/build.sh#L196

nickboldt:
> This message happens when the builder is trying to cleanup images from local cache https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/dependencies/che-plugin-registry/build.sh#L196

Well, I guess we could pipe output to `2>&1 >/dev/null` to suppress it but if it's just cleanup of image that was aleady cleaned up, all good.

SDawley:
Is this just to double check that the download script worked correctly? Seems like the same functionality.

nickboldt:
> Is this just to double check that the download script worked correctly? Seems like the same functionality.

I believe it's to compare the version of vscode required by the .vsix file to the version available from the code-rhel8 container.

See https://github.com/redhat-developer/devspaces/pull/975/files#diff-2a1490f6d73c0ae992afa03a522e1a6467190867ac28cc07016bac8fa3efddaeR50-R61

svor:
@SDawley you are right, the logic is the same. The difference is that in download_vsix.sh we check metadatas from https://open-vsx.org/api before downloading extensions
but in check_compatibility.sh we check package.json files from already downloaded vsix files that are going to be published into the embedded plugin registry.

nickboldt:
So would it make sense to do the check in this PR immediately after the download is done in `download_vsix.sh` ? that way the build would break sooner, if I understand the flow correctly.

svor:
@nickboldt  @SDawley the verification was moved into download_vsix.sh file. 
[Uploading screencast-che-dogfooding.apps.che-dev.x6e0.p1.openshiftapps.com-2023.07.07-17_55_49.webm…]()

SDawley (approved):

dependencies/che-plugin-registry/build/scripts/download_vsix.sh
```diff
@@ -139,5 +163,27 @@ for i in $(seq 0 "$((numberOfExtensions - 1))"); do
...
+        # check if the extension's engine version is compatible with the code version
+        # if the extension's engine version is ahead of the code version, check a next version of the extension
+        if [[  "$vscodeEngineVersion" = "$(echo -e "$vscodeEngineVersion\n$codeVersion" | sort -V | head -n1)" ]]; then
+            #VS Code version >= Engine version, can proceed."
+            echo -e "${GREEN}${EMOJI_PASS}${RESETSTYLE} compatible."
+        else 
+            echo -e "Extension requires a newer engine version than Che Code version ($codeVersion)."
+            echo -e "${RED}${EMOJI_FAIL}${RESETSTYLE} Test failed!"
+            exit 1
+        fi
```
ArvinB, line 187: I just ran into an issue here, and I thought it would be easier and more subtle to just add it in as a comment here for you @SDawley 
...

I have an extension that only has a `pre-release`. Specifically [Zowe.secrets-for-kubernetes](https://open-vsx.org/api/Zowe/secrets-for-kubernetes). This script breaks for this use-case versus falling through to the next extension. The `resultedVersion` variable never gets populated and thus error's out. Any chance you guys can correct this for an upcoming release?

PR #980, opened 2023-07-10
Fix arguments in updatePNCBuildConfigs.sh

SDawley:
The arguments for cloning have changed. 
--scm-revision is now --scmRevision
--product-version-id is gone

@nickboldt there is now a --profile option but I'm still researching that so for now I've just taken out product version id and fixed scm revision.


nickboldt (approved):
assuming you tested this, +1.

PR #984, opened 2023-07-17
Update download_vsix.sh to retry getting metadata if it fails because…

SDawley:
… the internet is unreliable.

### DIFF
--- a/dependencies/che-plugin-registry/build/scripts/download_vsix.sh
+++ b/dependencies/che-plugin-registry/build/scripts/download_vsix.sh
@@ -57,6 +57,11 @@ function getMetadata(){
         vsixMetadata=$(curl -sLS "https://open-vsx.org/api/${vsixName}/${key}")
         if [[ $(echo "${vsixMetadata}" | jq -r ".error") != null ]]; then
             echo "Attempt $j/5: Error while getting metadata for ${vsixFullName} version ${key}"
+
+            if [[ $j = 5 ]]; then
+                echo "[ERROR] Maximum attempts reached."
+                exit 1
+            fi
             continue
         else
             break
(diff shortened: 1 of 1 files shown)


dependencies/che-plugin-registry/build/scripts/download_vsix.sh
```diff
@@ -46,6 +47,23 @@ function initTest() {
...
+
+    # check there is no error field in the metadata and retry if there is
+    for j in 1 2 3 4 5
+    do
+        vsixMetadata=$(curl -sLS "https://open-vsx.org/api/${vsixName}/${key}")
+        if [[ $(echo "${vsixMetadata}" | jq -r ".error") != null ]]; then
+            echo "Attempt $j/5: Error while getting metadata for ${vsixFullName} version ${key}"
+            continue
+        else
+            break
```
nickboldt, line 62: on exit of this function, should ` vsixMetadata` be reset to "" so we don't accidentally use a previous value? or will `curl -sLS "https://open-vsx.org/api/${vsixName}/${key}"` ALWAYS get a value that overrides the previous global value?

nickboldt: also if this fails after 5 attempts... what's the resulting value of `vsixMetadata` that will be used in subsequent commands?

nickboldt: sounds like a plan. 

https://media.tenor.com/wHA-5f_wIWYAAAAM/shaun-the-sheep-thumbs-up.gif

dependencies/che-plugin-registry/build/scripts/download_vsix.sh
```diff
@@ -57,6 +57,11 @@ function getMetadata(){
         vsixMetadata=$(curl -sLS "https://open-vsx.org/api/${vsixName}/${key}")
         if [[ $(echo "${vsixMetadata}" | jq -r ".error") != null ]]; then
             echo "Attempt $j/5: Error while getting metadata for ${vsixFullName} version ${key}"
+
+            if [[ $j = 5 ]]; then
+                echo "[ERROR] Maximum attempts reached."
```
nickboldt, line 62:
```suggestion
                echo "[ERROR] Maximum of 5 attempts reached - must exit!"
```

nickboldt (approved):

PR #988, opened 2023-07-20
Fix: check against tryAgain the second time since checking against tr…

SDawley:
…y results in attempting to check in the same asset twice even if it succeeds.


nickboldt:
thanks! good fix

PR #999, opened 2023-08-28
Update uploadAssetsToGHRelease.sh to delete tags

SDawley:
Add a line to specifically find and remove the associated tag and maybe fix dsc.


nickboldt:
This (probably) does the same as line 22, where I push a tag deletion

SDawley:
Line 22?
I don't see anything at 22 or 222 that looks like tag deletion, if you mean https://github.com/redhat-developer/devspaces/blob/b78c56a42f73d53c19a1bdb7680de3f23f250857/product/uploadAssetsToGHRelease.sh#L120C5-L120C61
release delete and tag delete are different

It uses the word tag in the command description, but I think it's just the name of the release, and that's why the tag isn't getting removed.
I ran tag --delete locally and it worked.

SDawley:
OK that is a delete but for some reason it was not working, because the release would be removed but the tag remained...

SDawley:
Update: we have 2 green builds in a row I think it works.

PR #1003, opened 2023-09-05
chore:wrap DS version in quotes

svor:
### What does this PR do?
The value of DS version should be wrapped in quotes to parse it correct by jq.

Some scripts use `jq -r '.Version' "${jobconfigjson}"` to get version of DS, without quotes we get 3.1 instead of 3.10.

Fixes the plugin registry build

### DIFF
--- a/dependencies/job-config.json
+++ b/dependencies/job-config.json
@@ -1,6 +1,6 @@
 {
   "Purpose": "To set configurations of jenkins jobs including upstream branches, dependencies and disabled status",
-  "Version": 3.10,
+  "Version": "3.10",
   "Jobs": {
     "code": {
       "3.8": {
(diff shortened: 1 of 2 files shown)


SDawley:
@svor could you also make the corresponding change to https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/product/updateVersionAndRegistryTags.sh to include the quotations so it doesn't disappear next release when I run the script?

SDawley:
That's the function that we use to update every field.
Line 203 is where the version field specifically gets updated
https://github.com/redhat-developer/devspaces/blob/84aa00581de7992e73e96c8cefaf084c253a1c63/product/updateVersionAndRegistryTags.sh#L203

svor:
> That's the function that we use to update every field. Line 203 is where the version field specifically gets updated
> 
> https://github.com/redhat-developer/devspaces/blob/84aa00581de7992e73e96c8cefaf084c253a1c63/product/updateVersionAndRegistryTags.sh#L203

Replacing only .Version field

SDawley (approved):

PR #1011, opened 2023-09-20
fix: Fix cvs version prev in job-config.json.

SDawley:
Fix csv version prev in job-config.json. 
Fix extra comment in pluginregistry yaml (and made note to investigate why that happens). 
Take slack notification update out of update script since it is no longer relevant.


nickboldt (approved):

nickboldt:
this looks outdated, but I'll let @SDawley rebase or close it.

PR #1021, opened 2023-11-01
Fix: CRW-4976 add handling for different downstream branches.

SDawley:
https://issues.redhat.com/browse/CRW-4976

@nickboldt 
If this change looks good I can add something similar to the remaining get-sources.sh files.
Will also, obviously, have to make corresponding changes to the get-sources-rhpkg job in jenkins


nickboldt (approved):
I thought I'd already done something like this but maybe I never pushed it / left in PR limbo. Change looks reasonable.

PR #1032, opened 2023-12-12
Remove Camel devfile

apupier:
The devfile is unmaintained and still based on old versions of Fuse. Fuse will be reaching EOL anyway soon.
New devfiles based on Red hat build of Camel would better to be added (later and if some time to do it is allocated for that)

### DIFF
--- a/dependencies/che-devfile-registry/devfiles-extra/02_java11-maven-fuse/meta.yaml
+++ b/dependencies/che-devfile-registry/devfiles-extra/02_java11-maven-fuse/meta.yaml
@@ -1,7 +1,0 @@
----
-displayName: Red Hat Fuse
-description: Red Hat Fuse stack with OpenJDK 11 and Maven 3.6.3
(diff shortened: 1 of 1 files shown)


svor (approved):
LGTM

nickboldt:
Perhaps we want to remove everything in the "extras" folder?

https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/dependencies/che-devfile-registry/devfiles-extra/README.adoc

Or maybe just update the above README to remove the ones we definitely don't support anymore?

svor:
> Perhaps we want to remove everything in the "[extra](https://github.com/redhat-developer/devspaces/tree/devspaces-3-rhel-8/dependencies/che-devfile-registry/devfiles-extra)" folder?
> 
> Or maybe just update the [README](https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/dependencies/che-devfile-registry/devfiles-extra/README.adoc) in that folder to remove the ones we definitely don't support anymore?

@nickboldt I'll create an issue for this

PR #1034, opened 2023-12-12
fix: update parent folders for java based devfiles

svor:
### What does this PR do?
Rename parent folders of java based devfiles by changing `java11` prefix with `java`

https://issues.redhat.com/browse/CRW-4948


nickboldt (approved):
+1 but we could also just rename to `java` instead of `java17` so we don't embed the version at all

dmytro-ndp:
I would also suggest not to mention Java version in sample directory, to simplify support.

PR #1046, opened 2024-02-21
CRW-5708 Updating leftover instances of http to https.

SDawley:
### What does this PR do?
Some of the scripts that we use frequently but rarely update still had http when there are now https urls available.

https://issues.redhat.com/browse/CRW-5708


dmytro-ndp (approved):
looks good to merge

PR #1053, opened 2024-03-04
Updating copyIIBsToQuay to find the correct DWO version. 

SDawley:
### What does this PR do?
See title.
Also fixing extra slashes that was throwing an error in newer versions of grep.

The copyIIBsToQuay script defaults to devspaces-3-rhel-8 if its run from a different directory or outside of a devspaces-3.##-rhel-8 branch. The copy IIBs to quay job doesn't download the whole repo and just grabs individual scripts it needs. This meant that for release branches it was using the 3.12 version but the base branch data for everything else, and was looking for a DWO bundle that doesn't exist yet, so it was failing.

Unfortunately this script set up all the defaults before parsing arguments, so it took a large reorganization to fix this. I tested it locally and it works, if anyone else wants to double check run this:

`/product/copyIIBsToQuay.sh -v -t 3.12`

It will take a LONG time to run to completion but it will post the versions/branches fairly early on.

https://issues.redhat.com/browse/CRW-5774


mkuznyetsov (approved):
seems good!
if anything, do files need copyright header years updated?

SDawley:
> seems good! if anything, do files need copyright header years updated?

I misunderstood when I answered in standup, yeah I can make another PR updating the copyright dates.

PR #1062, opened 2024-03-29
feat: set fixed versions for vsix extensions; add script to update ve…

svor:
…rsions

### What does this PR do?
- Set fixed versions for vsix extensions that are going to e included into the embedded plugin registry;
- Add a script that can update extensions' versions to the latest compatible from opensx.org

https://issues.redhat.com/browse/CRW-4116

### DIFF
--- a/dependencies/che-plugin-registry/build/scripts/download_vsix.sh
+++ b/dependencies/che-plugin-registry/build/scripts/download_vsix.sh
@@ -8,8 +8,7 @@ set -o pipefail
 downloadVsix=1
 openvsxJson="/openvsx-server/openvsx-sync.json"
 
-usage()
-{
+usage() {
     echo "Usage: $0 -b devspaces-3.y-rhel-8 -j /path/to/openvsx-sync.json --no-download
 
 All arguments are optional.
@@ -22,13 +21,19 @@ All arguments are optional.
 
 # commandline args
 while [[ "$#" -gt 0 ]]; do
-  case $1 in
-    '-b'|'--branch') scriptBranch="$2"; shift 1;;
-    '-j'|'--json') openvsxJson="$2"; shift 1;;
-    '--no-download') downloadVsix=0;;
-    '-h'|'--help') usage;;
-  esac
-  shift 1
+    case $1 in
+    '-b' | '--branch')
+        scriptBranch="$2"
+        shift 1
+        ;;
+    '-j' | '--json')
+        openvsxJson="$2"
+        shift 1
+        ;;
+    '--no-download') downloadVsix=0 ;;
+    '-h' | '--help') usage ;;
+    esac
+    shift 1
 done
 
 RED="\e[31m"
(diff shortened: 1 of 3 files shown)


svor:
With this PR we are going to set fixed versions for all extensions (currently they are latest versions that are compatible with latest che-code version), they won't be updated automatically unless someone updates them by hands or running `update_extensions_versions.sh` script. It was done to support our own build of extensions with fixed versions in https://github.com/redhat-developer/devspaces-vscode-extensions/tree/devspaces-3-rhel-8 (to be done by https://issues.redhat.com/browse/CRW-4859).

What it gives:
- We know for sure which version is going to be used in the embedded plugin registry and we will build this version in https://github.com/redhat-developer/devspaces-vscode-extensions/tree/devspaces-3-rhel-8, if it is not possible to build on our side - we can change the version. I'd say it's a positive moment
- When new version of some extension is published on openvsx.org, it won't be used by Dev Spaces automatically. To use it we'd need to update the version in openvsx-sync.json. (It's an additional work and probably we'll have new issues from customers "Update extension X to the latest version")
- When we'll switch to use extensions [that are building on our side](https://github.com/redhat-developer/devspaces-vscode-extensions/tree/devspaces-3-rhel-8) and don't download vsix from openvsx.org , we'd also need to support the build of this new extension's version in https://github.com/redhat-developer/devspaces-vscode-extensions/tree/devspaces-3-rhel-8.

cc @ibuziuk @SDawley @mkuznyetsov

dependencies/che-plugin-registry/build/scripts/download_vsix.sh
```diff
@@ -70,11 +74,10 @@ getMetadata(){
 }
 
 versionsPage=""
-getVersions(){
+getVersions() {
     vsixName=$1
     # check the versions page is empty and retry if it is
-    for j in 1 2 3 4 5
-    do
+    for j in 1 2 3 4 5; do
```
ibuziuk, line 80: why 5 ?

SDawley:
The  update script for the devspaces-vscode-extensions half of things will have to be updated so it doesn't run the download_vsix script.

dependencies/che-plugin-registry/build/scripts/download_vsix.sh
```diff
@@ -22,13 +21,19 @@ All arguments are optional.
...
+    case $1 in
+    '-b' | '--branch')
+        scriptBranch="$2"
+        shift 1
+        ;;
+    '-j' | '--json')
+        openvsxJson="$2"
+        shift 1
+        ;;
+    '--no-download') downloadVsix=0 ;;
```
SDawley, line 33: If the update_extension_versions script is run manually do we still need the --no-download option? As I recall that was added so I could generate the versions for the [plugin-config.json](https://github.com/redhat-developer/devspaces-vscode-extensions/blob/devspaces-3-rhel-8/plugin-config.json)

dependencies/che-plugin-registry/build/scripts/download_vsix.sh
svor: sometimes it fails to get extension's metadata from the openvsx because the internet is unreliable or problem with the openvsx.org server, here we retry to read it

dependencies/che-plugin-registry/build/scripts/download_vsix.sh
svor: I think it's not needed, all versions will be in openvsx-sync.json. But we can't use these versions to get sources of extensions. Extension's version and source's revision is not identical :(

SDawley: Yeah I remember that being a pain to try and automate. I haven't looked at the script in a while but as I recall the process was to get the version from openvsx_sync.json, check if there was a tag which included that version, otherwise use the version to get the release date and find the previous commit.

90% sure that's what I did here [update-from-ovsx.sh](https://github.com/redhat-developer/devspaces-vscode-extensions/blob/devspaces-3-rhel-8/build/update-from-ovsx.sh)

svor: Good to know that we have a mechanism to get appropriate sources :+1:
So, @SDawley are you good with current approach to have fixed versions in openvsx-sync.json or maybe you have any concerns/suggestions?
Maybe we should extend openvsx-sync.json and add one more field for each extension that will provide reference to the github sources? In this case we won't need [update-from-ovsx.sh](https://github.com/redhat-developer/devspaces-vscode-extensions/blob/devspaces-3-rhel-8/build/update-from-ovsx.sh) at all, I think.

svor: I've removed `--no-download` parameter from the script

svor:
New GH action was added to check new vsix versions and create a PR if new versions exist. This action should be executed on the 1st day of each month

svor:
@SDawley could you please have another look at this PR. If there are no concerns I'd like to merge this PR and continue work on https://github.com/redhat-developer/devspaces-vscode-extensions part

ibuziuk (approved):

SDawley (approved):

PR #1082, opened 2024-07-04
skip external images update

svor:
### What does this PR do?
Since editors definitions are located in the operator side, we don't need to update references in plugin registry entrypoint.sh   

https://issues.redhat.com/browse/CRW-6794


dmytro-ndp (approved):
changes are checked in DS 3.15.0 https://github.com/redhat-developer/devspaces/pull/1081#pullrequestreview-2159326250

PR #1094, opened 2024-08-22
Don't use plugin-registry-generator tool in the plugin registry build process

svor:
### What does this PR do?
Removes plugin-registry-generator tool from the build process

- https://issues.redhat.com/browse/CRW-6884
- https://issues.redhat.com/browse/CRW-6932


SDawley (approved):
Looks good. It still builds so I'm all for removing unneeded parts.

PR #1099, opened 2024-09-06
Update openvsx config; remove /v3/plugins from the container

svor:
### What does this PR do?
- We don't need /v3/plugins/ folder anymore
- Add version of openvsx registry into server config

Dependent PR: https://github.com/eclipse-che/che-operator/pull/1894 

Related to https://github.com/eclipse-che/che/issues/23074


SDawley (approved):
lgtm

PR #1118, opened 2024-11-15
chore: create /var/run/postgresql folder before initialization

svor:
### What does this PR do?
Fixes plugin reg build by creating /var/run/postgresql folder before db initialization

https://issues.redhat.com/browse/CRW-7609


SDawley (approved):
Thanks for the fix!

